PR #41, opened 2018-09-21
Allow the user to override default days parameter on the command line

WebSpanner:


WebSpanner:
@freescout-helpdesk For testing yes, but mainly for initial import. The cron behaviour won't be affected as the default is still 3 days and the new option is optional.

WebSpanner:
@freescout-helpdesk Thanks!

PR #95, opened 2018-11-07
Fix #91

### LINKED ISSUE #91: Send auto-reply on every new conversation
Currenlty, auto-replies are sent once per day per customer.
This is not good, one auto-reply should be sent once per conversation to notify customers that their requests are received properly.

Probably, a setting could be added to enable or disable this.

If you are using Freescout as ticketing help desk, when the same user is creating multiple conversations, a receive acknowledge should be sent every time.

guestisp:
Fix #91 - Implement feature to always send an auto reply on new conversations.

Fix disabling send_auto_reply. unchecked checkboxes are not sent through the form and thus not avaialable in $request


freescout-helpdesk:
Thanks. Checking this. It would be good to change this option a little.

> Always send auto reply on new conversations
> (When a customer emails this mailbox with a new conversation, always send an auto reply to the customer regardless the 24-hours limit.)

We are aware that auto replies are sent once in 24 hours, but those who just installed the app and see this option will be confused.

Better to have a boolean `auto_reply_time_limit` option labeled something like `Send Once in 24 Hours`. This will be much shorter and there will be no confusion. Let's keep it `false` by default, those who want to have the limit will enable this option. Besides if in the future somebody will need to change 24 hours limit to 3 hours for example, we will be able to use this field and just convert into integer.

Also mentioning 24-hour period should be removed from the `Enable Auto Reply`'s hint.

guestisp:
Why not changing the option directly to an integer with default as 24?

Setting it to 0 will always send the reply, Setting it to a value created than 0 will send one email every X hours

freescout-helpdesk:
Yes. This also will be ok. Just small notice will be needed to explain, that "0" means - always send auto reply.

guestisp:
IMHO, keeping the 24-hours hard-limit for any new conversations is non-sense. When customers starts a new conversation, they should be notified every time, as an acknowledge.

Thus, why not simply remove these check and set auto-reply on/off without the 24 hours limit ?
Currently, autoreply is sent only on new conversations. I don't see any reasons to not notify customers that their request was received properly and would be handled shortly.

It would be much easier, no database migration or whatever, just remove the 24-hours limit.

Let me know what do you think so that I can start working on it asap

guestisp:
> In the migration file name it would be good to mention table name: 
> add_auto_reply_time_limit_column_to_mailboxes_table.php

Ok.

freescout-helpdesk:
Agreed. If customer sends a first message to support and receives auto reply, when he sends a second request and does not receive an auro reply, he would think that something is broken and that his request has not been received. So let's just disable this 24h limit.

guestisp:
Ok. give me a couple of hours, i'm stuck with another work right now

PR #98, opened 2018-11-08
Add support for conversation number format

### LINKED ISSUE #97: Random ticket number
Ticket number should be created randomly (or with something not easily guessable)

guestisp:
Fix #97

Add support for conversation number format (sequential, random_numeric, random_alphanumeric)
Add support for setting the next conversation number in sequential mode

### DIFF
--- a/app/Conversation.php
+++ b/app/Conversation.php
@@ -5,6 +5,7 @@ namespace App;
 use App\Events\ConversationCustomerChanged;
 use App\ConversationFolder;
 use App\Folder;
+use App\Option;
 use App\Thread;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\Input;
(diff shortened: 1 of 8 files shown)


freescout-helpdesk:
Thanks. Took a quick look. Few questions:

1. Conversation::max('number') is not working when `numbder` is string?

1. What for  `doctrine/dbal` is used? It will make application incompatible with PHP 7.0

guestisp:
> Conversation::max('number') is not working when numbder is string?
I don't think it would work, as it's using `MAX` from mysql. I've tried by calling the same query manually over a varchar column and it's returning zero. So I did that trick.
> What for doctrine/dbal is used? It will make application incompatible with PHP 7.0

dbal is used for the migration. Is not possible to alter the column type without dbal. If you try, artisan tell you to install dbal.

freescout-helpdesk:
Something like this can be used to change column type
```
public function up()
{
    \DB::statement('ALTER TABLE conversations MODIFY mycolumn  VARCHAR(7);');
}
public function down()
{
    \DB::statement('ALTER TABLE conversations MODIFY mycolumn INT(10) UNSIGNED;');
}
```

This
```
Conversation::where('number', $next_ticket)->exists();
```
can be a little faster than
```
Conversation::where('number', $next_ticket)->count();
```

guestisp:
Ok. Tomorrow i'll fix.
Are you able to make some test with and without existing data?

I did many test successfully but i would like to have an additional confirm.

Obvioulsly is not possibile to rollback after adding conversations with alphanumeric numbers but this should never happens

freescout-helpdesk:
One more thing - it would be better to have integer constants in App\Conversation instead of using strings `sequential`, etc.

This constant can also be passed to JS as a function argument:
```
initGeneralSettings({{ App\Conversation::NUMBER_FORMAT... }})
```

guestisp:
Another solution would be to use the table id in case of sequential numbering
With this, we can use max on id column and remove the ugly cast on "number" but imho noone should use an easy to guess sequential numbering....

guestisp:
You can't switch from random to sequential when conversations already exists

I don't think there would be a solution. If you have a conversation with number "AB45ZDG" How do you find the higher sequential number? You don't have numbers....

Probably we could block the format change If some conversations with a random scheme already exists

guestisp:
> Why do you think people should not use easy to guess sequential numbering?

For security. Is not a best practice to expose sequential IDs, that's why usually the table autoincrement should not be exposed to customers

freescout-helpdesk:
OK. So on the Settings page:
- If user chooses `Random...` format there should be some `text-danger` notice displayed explaining that it will be impossible to switch to sequential format afterwards.
- If there is at least one conversation with non-sequential number, `Random...` options should be disabled.

guestisp:
> If there is at least one conversation with non-sequential number, Random... options should be disabled.

The opposite: if there is at least one conversation with non-sequential number (how to check this?), sequential option should be disabled.

freescout-helpdesk:
> The opposite: if there is at least one conversation with non-sequential number (how to check this?), sequential option should be disabled.

Check if there is a conversation with number containing 7 chars.
In the `Next Conversation #` field allow to enter only values <= 999999

freescout-helpdesk:
See how it looks in Help Scout:


Sequential IDs are not exposed to the end users. So it there a real reason to implement random conversation numbers?

freescout-helpdesk:
> “When we built Help Scout, choosing to say ‘conversations’ instead of ‘tickets’ informed so many of our decisions, like simplifying the customer portal and removing ugly ticket numbers that are mostly there for computers to read.”
– Nick Francis, CEO

https://www.helpscout.net/blog/support-ticket-system/

guestisp:
Ok, if it is not exposed, there is no need to change the format and I can drop this PR.

I agree with the HelpScout quote, but how would be possible, for a customer, to keep track of their requests or how would be possible, for a support agent, to refer to a single conversation with a customer?
In example: "Look at request #XXXXXXX you made 2 days ago". If you don't have a tracking number, you won't have a reference to use.

freescout-helpdesk:
We are using Help Scout for several years and never faced such issue.

guestisp:
Ok. So, let's drop this PR ed fix the next ticket number

freescout-helpdesk:
Yes.

PR #102, opened 2018-11-09
Allow to set the next conversation number

guestisp:


freescout-helpdesk:
1. We should do `Option::remove` only if there is something in `$next_ticket`.

2. `Next Conversation #` should display current value from the option or current `max(number)+1`

4. If user sets `Next Conversation #` less than current max(number)+1, entered value should be set to max(number)+1

4. `{}` are needed for the IF to follow Laravel standarts ([exampple](https://github.com/laravel/laravel/blob/d53539b47b6d2a20d0c3d3257ad801bc72547c9d/app/Http/Middleware/RedirectIfAuthenticated.php)):
```
if ( !$next_ticket || Conversation::where('number', $next_ticket)->exists() )
    $next_ticket = Conversation::max('number') + 1;
}
```

guestisp:
> We should do Option::remove only if there is something in $next_ticket.
Ok.

> Next Conversation # should display current value from the option or current max(number)+1

I think this would make confusion. Because you can't know if the shown value is something you have manually set or is the next automatic number. In addition, if you open the setting page, you'll se a value (`current max(number)+1`) Then a new conversation is created when you have the setting page opened. When you press save, you are setting a wrong number because the `current max(number)+1` is not more correct but it was set in the field when you opened the page

> If user sets Next Conversation # less than current max(number)+1, entered value should be set to max(number)+1

Ok.

> {} are needed for the IF to follow Laravel standarts (exampple):

Ok

guestisp:
I've pushed some changed that should fix what you told me. 
but I really don't like the point "2", it makes confusion and it's prone to errors

freescout-helpdesk:
> Next Conversation # should display current value from the option or current max(number)+1

This is how Help Scout works. When I see empty field and I don't not know what is the current number, I feel confused.

> When you press save, you are setting a wrong number because the current max(number)+1 is not more correct but it was set in the field when you opened the page

In self::creating `$next_ticket = Conversation::max('number') + 1;` must be retrieved every time anyway, even if $next_ticket is set, because real max(number) can be higher than value set in `$next_ticket`.

freescout-helpdesk:
Perfect. Will try to test tomorrow.

guestisp:
Thank you

PR #105, opened 2018-11-09
Add tracking of opened emails

guestisp:
This is a WIP. gmail doesn't load the remote image used for tracking openings


freescout-helpdesk:
Is it ready for testing?

guestisp:
Yes. Maybe you can tell me why opening is not tracked with Gmail even if the tracking image is present

I will look at this tomorrow

freescout-helpdesk:
1. `ConversationsController.php` requires user to be authenticated (see it's `__construct`). You can use `PublicController.php`
1. Add name to the route in web.php: `->name('open_tracking.set_read');`
1. URL in `reply_fancy.blade.php` should be generated with `route('open_tracking.set_read')`

guestisp:
So, what about to use a new controller, like `NotificationController` that would be used for any kind of these notification ? HelpScout has a route called `/notification` thus I'm thinking that would be other notification in the future.

freescout-helpdesk:
Just use `PublicController.php` for now. Later will see.

We call `Notifications` emails sent to users. Emails sent to customers in FreeScout are called `replies to customers`

guestisp:
Ok, I did your suggestions. Now is working properly.

guestisp:
This also fix #108. It was a trivial change on the same file, I think there is no need to create another PR

guestisp:
Thank you

Issue #108, opened 2018-11-09
Show thread date when hovering


freescout-helpdesk:
Where the conversation date is displayed, if thread date is displayed as "3 days ago" there should be a tooltip showing the full date "Oct, 17" or "Oct 17, 2017".

freescout-helpdesk:
Implemented

PR #127, opened 2018-11-13
Show thread date when hovering #108

labomatik:
tooltip showing the full date


freescout-helpdesk:
Thanks. Please use `User::dateFormat($date, $format)` function to display date.
https://github.com/freescout-helpdesk/freescout/wiki/Development-Guide

PR #129, opened 2018-11-14
Conversations bulk actions

Junker:


freescout-helpdesk:
Thanks. Will check it soon.
Can you use main ajax() method in ConversationsController for ajax actions instead of ajaxBulk().

Junker:
IMHO keep all ajax requests in one function is wrong, this one function will grow and grow. 
I think need to make refactoring, and add separated route for every action.

freescout-helpdesk:
No worries. Modular architechture will not allow it to grow too much.

freescout-helpdesk:
When using search:

> Undefined variable: mailbox (View: /resources/views/conversations/partials/bulk_actions.blade.php)

freescout-helpdesk:
Will fix it.

PR #149, opened 2018-11-24
add history threads for restore/delete conversation action

Junker:


freescout-helpdesk:
Thanks. Just no need to mark thread as deleted:
```
$thread->state = Thread::STATE_DELETED;
```
Conversation can be deleted. Thread can not be deleted.

Junker:
fixed

PR #151, opened 2018-11-24
Fix "Class 'App\Console\Commands\Exception' not found" error

dryabov:


dryabov:
Hmm, I edited master branch on github, most likely I missed select of branch for pull request, and github set it to default branch (dist). Console git is more predictable than this web interface.

PR #152, opened 2018-11-24
typo

dryabov:
flollowing -> following


dryabov:
There are many changes after that, I think it's simpler for you to commit this change (if it is not covered by recent "Improve text" request with many fixes).

PR #160, opened 2018-11-26
Long press instead of checkboxes in mobile version

Junker:


freescout-helpdesk:
Looks good. Just attachment icon should be made visible when needed next to the conversation title:


If you will have difficulties doing this, just leave it, we will fix this later.

Junker:
fixed

PR #172, opened 2018-12-06
Add nullable to in_password field in mailboxes table

guestisp:


guestisp:
Not in mine:
```
    public function up()
    {
        if (version_compare(config('app.version'), '1.0.6', '<')) {
            Schema::table('mailboxes', function (Blueprint $table) {
                $table->string('in_password', 512)->change();
            });

            foreach (\App\Mailbox::whereNotNull('in_password')->get() as $Mailbox) {
                $Mailbox->in_password = $Mailbox->getOriginal('in_password');
                $Mailbox->save();
            }
        }
    }
```

and even not in yours: https://github.com/freescout-helpdesk/freescout/blob/master/database/migrations/2018_11_13_143000_encrypt_mailbox_password.php#L18

guestisp:
This because the `down` migration will remove the nullable:

```
            Schema::table('mailboxes', function (Blueprint $table) {
                $table->string('in_password')->change();
            });
```

during development i had to rollback multiple times, this removing the `NULL` flag from the table.

This PR request should fix all users that rolledback my previous migration `2018_11_13_143000_encrypt_mailbox_password`

guestisp:
Usually, changing an old migration is a bad practise

guestisp:
> In this case it's ok, if someone managed to do a rollback, it means he/she is a developer and will be able to deal with this.

Fixed

PR #184, opened 2018-12-14
Bounce detection improvement

guestisp:
Small improvement to bounce detection. Quick & Dirty fix to detect AWS bounces and Exim bounces

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -242,7 +242,7 @@ class FetchEmails extends Command
                     if ($message->hasAttachments()) {
                         foreach ($attachments as $attachment) {
                             if (!empty(Attachment::$types[$attachment->getType()]) && Attachment::$types[$attachment->getType()] == Attachment::TYPE_MESSAGE) {
-                                if (in_array(strtoupper($attachment->getName()), ['RFC822', 'DELIVERY-STATUS', 'DELIVERY-STATUS-NOTIFICATION', 'UNDELIVERED-MESSAGE']) 
+                                    if (in_array(strtoupper($attachment->getName()), ['RFC822', 'DELIVERY-STATUS', 'DELIVERY-STATUS-NOTIFICATION', 'UNDELIVERED-MESSAGE'])
                                         || preg_match('/delivery-status/', strtolower($attachment->content_type))) {
                                     $is_bounce = true;
                                     $bounce_attachment = $attachment;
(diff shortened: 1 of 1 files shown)


freescout-helpdesk:
Conversation will be created always, it allows support agents understand that their email to customer has not been delivered.

Bounce detection is used to avoid sending auto replies, to avoid infinite loops: https://github.com/freescout-helpdesk/freescout/blob/master/app/Listeners/SendAutoReply.php#L34

guestisp:
Uhm, maybe we can move all bouces to a dedicated folder (something like the Spam folder) ?

This to avoid mixing customers requests from bounces. We have some transactional mail sent with a freescout mailbox (we are not sending these email via freescout, only the From is set to a freescout mailbox). All bounces are triggering a new conversation but in this case there is no need for a conversation, or at least, there is no need to flood the mailbox primary folder with bounces.

A dedicated folder and an autoreply block would be ok.

freescout-helpdesk:
> maybe we can move all bouces to a dedicated folder (something like the Spam folder)

Not in the core.

guestisp:
> Not in the core.

Why not ? Users can't automate the move. Why not adding a configurable option asking how to manage a bounce? Maybe: "Silenty delete them", "Move them to bounce folder", "Do nothing"

freescout-helpdesk:
You can create a feature request here https://freescout.net/request-feature/ to see if anybody needs it.
For now feel free to implement it as a custom module.

guestisp:
Any guide on modules development?

freescout-helpdesk:
See existing modules and:

- https://nwidart.com/laravel-modules/v2/introduction
- https://github.com/tormjens/eventy

guestisp:
Let me check, but I think a new filter should be called on bounce receipt. By using a custom module to just skip conversation in case of bounce seems a little bit overkill to me. Tons of code (mostly for module scheleton) when just 2 line of code are needed (a simple IF)

guestisp:
I've created a module for bounce management. There is only one missing thing that I don't know how to fix: I need to skip the `$this->saveCustomerThread` and `$this->saveUserThread` in `FetchEmails`

I've thought to create a filter, a sort of `preHook`. Do you have any suggestion on it's name ?

freescout-helpdesk:
> I've thought to create a filter, a sort of preHook. Do you have any suggestion on it's name ?

Something like `fetch_emails.should_save_thread`

PR #188, opened 2018-12-17
Add "fetch_emails.should_save_thread" hook

guestisp:
As per #184 here we add a new hook to block thread creation (needed by module developers)


guestisp:
As wrote on Eventy docs, any parameter following the filter name is sent as function argument. As some module could modify the $hook value (or just returning `false` to block the conversation creation) i've used the whole $hook as second parameter

If noone is listening to this hook, the original hook value is returned otherwise new values are returned from the module (or false to block the creation)

Doing as you are suggesting, the module can only return false

freescout-helpdesk:
`$hook` is not used in the code, so it is useless to change it via the filter. Filters mus return what they receive as a second parameter.

guestisp:
I'm planning to use $hook today.

freescout-helpdesk:
Just add another filter before `If`:
```
$data = \Eventy::filter('fetch_emails.data_to_save', $data);
```
`$hook` renamed into `$data`

guestisp:
With this you are forcing to use 2 hooks when 1 should be enough

guestisp:
Something like this ?

guestisp:
Thank you

PR #218, opened 2019-01-02
Bounce detection improvements

guestisp:


freescout-helpdesk:
Where does this solution come from?
`Return-Path` can be set to `<>` by some mail server for a regular email too.

guestisp:
I've never seen, in about 15 years, a single mail with return path set as <> for something different than a bounce (or spam/bulk email)

freescout-helpdesk:
Existing bounce detection methods do not detect some of your incoming bounce emails?

guestisp:
In some case, yes, that's why there are some additional log in this PR

PR #232, opened 2019-01-10
Fix To address in customer notification emails

guestisp:
When creating a new conversation from the control panel, the "To" address is set as empty if there isn't any fullname stored for that address. This will fix the empty string sent in To header by replacing it with the customer email address

### DIFF
--- a/app/Jobs/SendReplyToCustomer.php
+++ b/app/Jobs/SendReplyToCustomer.php
@@ -112,7 +112,7 @@ class SendReplyToCustomer implements ShouldQueue
         // }
 
         try {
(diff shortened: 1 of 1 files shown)


freescout-helpdesk:
It works fine. No need to try to fix something which works, most likely eventually it will lead to problems.
Taking care of the name in "To" header is Laravel swiftmailer's task.

guestisp:
Not exactly, because who is setting an empty string is freescout, not Laravel

freescout-helpdesk:
It looks like nobody experience any problems with this.
It does looks like you are waisting too much our time with useless improvements and false bug reports.

PR #261, opened 2019-05-01
fetch mail schedule setting

Junker:
Some email services have limits for connections, and make IMAP connection every minute is wrong.


freescout-helpdesk:
Could you please:
1) Name variable as `fetch_schedule` and save as env file parameter: APP_FETCH_SCHEDULE. See how `alert_logs_period` is implemented.
2) In Manage > Mail Settings add `System Emails` header and `Fetching Emails` section below.

Junker:
done

Issue #327, opened 2019-09-11
Attachments Not Viewable


Speedy059:
I've noticed that if an attachment has "#" in it, you can't view it. The attachments directs to a page with: "Sorry, the page you are looking for could not be found.". I've checked on the server and the file is there, but the "#" character throws it off.

Reproduce Instructions: Send an email with an attachment and put "#" somewhere in the file name (EX: PO#4262254.pdf). Doesn't matter on the extension, can be .pdf,.jpg,etc.

freescout-helpdesk:
Thanks. It will need to be fixed.

freescout-helpdesk:
Has been fixed a year ago.

PR #331, opened 2019-09-17
Fix for attachments with # in name

TWithers:
This is in reference to #327 

Updated the file_name of the attachment to slugify the name: only lowercase alphanumeric and slashes. Should solve all issues with # or spaces in file names.

A better option would be to store and display file names as they are sent, but use the id or slug+id of the attachment as the url instead of the name of the attachment.


freescout-helpdesk:
Thanks. But could you please create a push request into **master** branch? https://github.com/freescout-helpdesk/freescout/wiki/Development-Guide#github-workflow

Pull requests can not be merged into **dist**.

TWithers:
Updated it. Thanks!

PR #433, opened 2020-01-18
Added German translation

TimoStramann:
German translation was added to pre-installation messages and system messages.
PHP7.3 doesn't cause a bug so reverted for compatibility


freescout-helpdesk:
Thanks. But please push only translations.

PR #438, opened 2020-01-20
Adding new filters to allow hooking menu items from third party modules

santibm:
I'm not so sure if I'm right with this. I'll explain, I'm doing a custom module and wanted to add menu items either in the main menu or inside admin menu: 

Screen Shot 2020-01-20 at 12 03 22

So I added two new filter positions to the blade template


freescout-helpdesk:
Could you please rename hooks:

`admin.menu.after` >> `menu.manage.append`
`main.menu.after` >> `menu.append`

Issue #308, opened 2019-08-09
[Security] Unauthorised access to attachments


NdT3Development:
While testing Freescout (I am currently using version 1.2.6), I have found that the email attachments can be accessed without needing to be logged in by accessing the URL.

For example:
An email attachment with the name `image001.png` will be uploaded to the server. This uploaded file should only be accessible while a user is logged in however, even after logging out, I can still access the image through the URL (Example: https://REDACTED/storage/attachment/5/2/1/image001.png)

This is a security risk as some businesses deal with confidential information which should only be accessible to staff and not unauthorised users on the internet.

frommMoritz:
You could use a randomly generated UUID. For example: [https://mail.example.com/storage/attachment/8/5/856b6674-4987-4794-8d8a-6f7d8c7d7af8/filename.extension](https://mail.example.com/storage/attachment/8/5/856b6674-4987-4794-8d8a-6f7d8c7d7af8/filename.extension). This Approach would preserve the original filename but also prevent the guessing of common filenames such as `document.pdf`, `image001.png` or `logo.jpeg`. I think it should be an option, if a user must be actually logged in to view the attachment or if the specific path is enough. In our use case (mostly as a great multi-user mail-system) it is helpful having the ability to access documents on the mobile where not everyone is logged in.

But I see that it, depending on the context, it is not advisable to allow unauthorized access.

This libary could be used: https://github.com/ramsey/uuid/

Qrious:
We've used a different solution that is relatively simple to implement. Thought i'd share it, in case anyone is interested. If @freescout-helpdesk accepts this method, i'm willing to make a PR on the docs.  We use the [nginx auth-request module](http://nginx.org/en/docs/http/ngx_http_auth_request_module.html) (requires apt-get install nginx-full on ubuntu) to authenticate the download request. Basically this module allows you to make an 
`auth-request` for a location in nginx.

> The ngx_http_auth_request_module module (1.5.4+) implements client authorization based on the result of a subrequest. If the subrequest returns a 2xx response code, the access is allowed. If it returns 401 or 403, the access is denied with the corresponding error code. Any other response code returned by the subrequest is considered an error.

This has as added advantage that the file to be downloaded does not have to pass through PHP, but instead can be delivered by Nginx like any other regular file (less memory usage, and buffer copying overhead).

Example configuration:
```
server {
   # Other config removed for brevity
    error_page 401 = @error401;

    location @error401 {
        return 302 https://example.com/login;
[...]

PR #486, opened 2020-02-16
Require an access token to download files.

Qrious:
Fixes https://github.com/freescout-helpdesk/freescout/issues/308


freescout-helpdesk:
Thanks. Implementation is in the master branch: https://github.com/freescout-helpdesk/freescout/issues/308#issuecomment-596070656

Issue #487, opened 2020-02-16
Select2 tokenization incorrect when pasting multiple email addresses in to/bcc/cc field.


Qrious:
As depicted in the gif below, currently multiple email addresses are added as a single tag when copy+pasting a string containg multiple email addresses seperated by any of the supported delimiters. For example pasting:
`test1@example.com, test2@example.com, test3@example.com` results in the following situation:
![current](https://user-images.githubusercontent.com/680327/74608258-dcc44200-50df-11ea-990c-f0f424fce1b0.gif)
In this case, multiple tags with a single email address each should be added.

freescout-helpdesk:
Merged.

PR #488, opened 2020-02-16
Fix tokenization of multiple emails on copy+paste of multiple emails

Qrious:
This fixes freescout-helpdesk/freescout#487 with copy+pasting a string containing multiple emailaddresses separated by any of the allowed tokens. Before this patch, all emailaddresses became a single tag.

Before:
current

After:
fix

I'm aware that the last item is not correctly tokenized. This is due to https://github.com/select2/select2/issues/5076. I'm looking at applying the workaround as well, but it did not immediately work.

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -2414,7 +2414,7 @@ function initCustomerSelector(input, custom_options)
 				    }
 				}
 			    // Check if select already has such option
-			    var data = this.$element.select2('data');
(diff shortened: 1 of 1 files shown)


Qrious:
I've tried to applied the workaround, however this results in new issues (tags being about 5 pixels wide, and becoming 'flacky' in terms of being correctly converted to tags). I suggest we do not apply the workaround.

PR #520, opened 2020-03-08
Add commandline arguments to freescout:create-user

nebulade:
This allows for automation of initial admin account creation together with the already supported `--no-interaction` option.
All options are optional and if not provided will be asked for.

This requirement came up while I was packaging freescout for Cloudron due to popular request.


freescout-helpdesk:
Please let us know when FreeScout will be added to the Cloudron. We will add info to the project Readme.

Issue #447, opened 2020-01-23
Option to send full conversation history to customers


freescout-helpdesk:
todo

joris85:
Great!
I have had multiple customers that asked where i was replying on :-)
Would be a nice feature to have the possibility to send the history.

Getting more and more enthousiastic about Freescout!

freescout-helpdesk:
Before https://github.com/freescout-helpdesk/freescout/releases/tag/1.4.2 release there was an issue with grouping emails for customers (Fixed In-Reply-To header in replies to customers). Most likely after 1.4.2 release your customer won't have this issue anymore.

joris85:
I do not understand the status of this. Is the option coming that the customer is seeing what they asked or the full history of the ticket maybe?

I get daily confused customers asking where my reply was for exactly

nextima:
Same here - our customers are getting super confused after out recent update of Freescout, when they suddenly don't get the full conversation in our replies.
The option for enabling the history in the replies is must have, tiredofit suggested the good solution in  #365 [here](https://github.com/freescout-helpdesk/freescout/issues/365#issuecomment-553916972)

freescout-helpdesk:
"Conversation history" dropdown should be added to the Settings --> Mail Settings (Emails to Customers):

- No history
- Last customer message
- Full history

joris85:
Hi Freescout,
Does help wanted mean that you guys do not work on this?
I really need this function so would like to know if somebody is working on this or not.
Can this be done by a module or does it need to be in the core?

freescout-helpdesk:
`help wanted` label - work which need to be done, by anyone.
This should be in the core.

joris85:
I am no programmer, but i love this feature a lot.
No idea if this is alowed with the freescout and opensource but i am willing to hire my developer to build this and share in the money with who is willing to.
ofcourse afterwards it is core of Freescout.

Is there somebody who wants this and is willing to pay a bit for the development?

Or can i donate to Freescout for this feature?

joris85:
Any news on this?
@freescout-helpdesk do you guys have a way that multiple people can donate for a specific feature and you built it? I will :-)

I understaand it is FREEscout but you guys need to get money for something and i a willing to donate for a feature like this. I just need this as fast as possible. If would never be implemented i need to leave freescout. This is a must for me.

fvdm:
Thanks for adding this feature!

[...]

PR #535, opened 2020-03-13
Add option to include the last message in notifications

fvdm:
This adds a third option `Include the last message` to Settings > Conversation history (#447)

In my tests the notification mails look the same as the current 'Send full conversation history' with only the new reply and the one before that included.

I wasn't sure if the code should be added before or after the `\Eventy::filter()`, but it doesn't seem to matter.

~Edit: the staff notifications are still the full list, I don't know where that can be altered.~

### DIFF
--- a/app/Jobs/SendNotificationToUsers.php
+++ b/app/Jobs/SendNotificationToUsers.php
@@ -62,11 +62,6 @@ class SendNotificationToUsers implements ShouldQueue
             return;
         }
 
-        // Include only the last two messages if configured
-        if (config('app.email_conv_history') == 'last') {
-          $this->threads = $this->threads->slice(0, 2);
-        }
-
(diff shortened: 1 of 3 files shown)


fvdm:
I found the user notifications fix

freescout-helpdesk:
`app.email_conv_history` option should not affect user notifications.

fvdm:
> `app.email_conv_history` option should not affect user notifications.

We need it for GDPR compliance. Our clients send us PII by mail which is okay to appear in the initial notification to the assignee. But there should be the least amount of copies as possible when people are added to the conversation.

fvdm:
Should I add a toggle 'Also apply to user notifications' that is off by default?

freescout-helpdesk:
> Should I add a toggle 'Also apply to user notifications' that is off by default?

For this purpose create a separate issue.

PR #568, opened 2020-04-10
Fix for reply on notification e-mail not assigned to original conversion

ChrisRAoW:
Mac mail (at least) removes elements that are set to display none when you're replying. Because the {#FS:****#} was in an hidden div it was removed, and freescout was not able anymore to relate the reply to the original conversation. Simply removing the display none fixes that.


freescout-helpdesk:
Thanks. Could you please submit pull request into `master` branch: https://github.com/freescout-helpdesk/freescout/wiki/Development-Guide#github-workflow

ChrisRAoW:
My bad, close in favor of  #576

Issue #574, opened 2020-04-12
Reply mails in same mailbox for full threading


nebulade:
Hi,

(sorry if this was already answered somewhere, but I didn't find an answer so far)

we at Cloudron currently use a shared mailbox via mail clients for support and are about to start using FreeScout, for a better support flow. So far we have setup mail clients to place the response mails from our side next to the incoming mails, which allows for nice threading in mail clients, without loosing history in case the mailbox is access directly.
From what I can tell FreeScout does not put sent out reply mails anywhere in the actual mailbox (as far as I can tell not even into `Sent`). I wonder what I miss here with regards to a more centric email flow and what the recommendations are around how to ensure the reply mails end up also in the same mailbox. 

I did try to use the auto-bcc feature to ensure the replies get put into the support mailbox, but that just means the message flow within FreeScout gets confused and lists the double mails there.

Anyone with a similar use-case or ideas how to get this working?

Thanks a lot,
Johannes

freescout-helpdesk:
Actual reply emails can be found in the `Sent` folder of the mail service provider which is used.

imc67:
> Actual reply emails can be found in the `Sent` folder of the mail service provider which is used.

After 4 month of use I now realized that sent mails from Freescout (using SMTP of the same IMAP  mailbox) is not stored in any folder of the email provider. Is that correct?

freescout-helpdesk:
Outgoing emails should be stored in the `Sent` folder of the email provider. If they are not there, check with your email provider why.

nebulade:
I was always under the impression, that the email clients would put message in the sent folder, and thus FreeScout would have place a copy int the sent folder, after a quick grep through the code, I don't seem to find any code for that. Maybe some providers also have server side code to ensure a copy is put into the sent folder?

nebulade:
Yes so all I can find is the information that the email client would put the message into the configured sent folder, usually clients like Thunderbird detect this folder automatically and thus there is no further configuration required. Since FreeScout is the MUA in this case, it is responsible to place a copy of the sent mail somewhere into the mailbox. Ideally it would be possible configure this imap folder, with regular `sent` being the default.
Furthermore it would be really nice, if one could configure to copy the sent mail next to the other mails for threading. Given that FreeScout can already be configured from which imap folder incoming mails are picked, it would also save the reply mails in those folders from where the original inquiry mail was picked from.

[...]

PR #577, opened 2020-04-15
Save reply email to mailbox

nebulade:
Part of #574

### DIFF
--- a/app/Jobs/SendReplyToCustomer.php
+++ b/app/Jobs/SendReplyToCustomer.php
@@ -11,6 +11,7 @@ use Illuminate\Foundation\Bus\Dispatchable;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Queue\SerializesModels;
 use Mail;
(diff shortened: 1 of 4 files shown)


app/Jobs/SendReplyToCustomer.php
```diff
@@ -223,6 +226,27 @@ public function handle()
             }
         }
 
+        // TODO copy reply to sent only if configured
+        $client = \MailHelper::getMailboxClient($mailbox);
+        $client->connect();
+
+        // TODO this needs fixing to be properly encoded and all
+        $content = "From: " . $mailbox->getMailFrom()['address'] . "\r\n"
```
nebulade, line 234: I am currently stuck here, since I am not a php/laravel developer. What is the correct way to construct a valid and properly escaped mail string in php?

nebulade:
Is there anything blocking this pull request? I am more than happy to fix/adjust things

freescout-helpdesk:
It will be checked in several days.

PR #596, opened 2020-04-30
Update NL translation

fvdm:
Consistently use 'u' instead of 'je' for the English 'you'. This is a bit more formal language, but also more universal.

Also added missing translations and reordered the JSON keys.


fvdm:
Ah shoot, I used the file from the live instance thinking the translator in the CP would write to this file, but it apparently didn't. I will update this PR in a moment.

fvdm:
Why has this PR been reverted? All the old inconsistent translations are back.

freescout-helpdesk:
We will check.

PR #600, opened 2020-05-02
Preserve the thread id if any for reply copies

nebulade:


nebulade:
Is there anything I should adjust for this to get accepted? Thanks.

freescout-helpdesk:
It will be checked soon.

PR #639, opened 2020-06-16
Replace brand logo with svg version to support hiDPI screens

nebulade:
Sorry this might be minor, but I thought might still be nice. I've created a brand icon svg which looks the same as the current png.
It was created with Inkscape and then I optimized it with https://jakearchibald.github.io/svgomg/


freescout-helpdesk:
Try this: [icon.zip](https://github.com/freescout-helpdesk/freescout/files/4789950/icon.zip)

nebulade:
I've updated the commit to include your proposed svg icon now

PR #665, opened 2020-07-10
Added conversation settings

Dirk-Kokx:
When sending a reply to a customer you can set if the history of the conversation should be send along.
This setting however is global, so to all conversations.

With this modification you now can change the setting per conversation.
For this a settings button is added in the topbar of a conversation.
For now this only includes history settings (an override to the global setting per conversation)


fvdm:
I would remove the advertising. Source code is not the place for that.

Dirk-Kokx:
You are right, fvdm. This was my first pull request ever and I thought it might help to point out where the differences were, rather then for advertising reasons. It appears to be redundant, and I updated the code.

freescout-helpdesk:
Could you please change the branch to **master** for the pull request.

Dirk-Kokx:
Sure, done!

freescout-helpdesk:
It looks like it does not work. Please recreate the pull request from your master branch.

PR #672, opened 2020-07-14
Added conversation settings

Dirk-Kokx:
For now this only includes history settings (an override to the global setting per conversation)

This is a new attempt from a fresh master fork.

### DIFF
--- a/app/Conversation.php
+++ b/app/Conversation.php
@@ -19,7 +19,7 @@ class Conversation extends Model
     use Rememberable;
     // This is obligatory.
     public $rememberCacheDriver = 'array';
-    
+
     /**
      * Max length of the preview.
      */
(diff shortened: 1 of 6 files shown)


freescout-helpdesk:
> With this modification you now can change the setting per conversation.
> For this a settings button is added in the topbar of a conversation.
> For now this only includes history settings (an override to the global setting per conversation)

We can't find a settings button in the topbar of a converrsation.

freescout-helpdesk:
Could you please:
- Change `email_conv_history` field to `$table->unsignedTinyInteger('email_conv_history')->default(0);` and use constants similar to Conversation::TYPE_ and create `getEmailConvHistoryName()` method for Conversation.
- Change "Save and close" to "Save".
- In the dropdown change "Use the default settings" to "Default (Send full conversation history)", where `Send full conversation history` - is the global setting value.
- Change modal title to "Conversation Settings".
- Move "Settings" to the very bottom of the dropdown.

Issue #612, opened 2020-05-16
Rescale large images for readability


fvdm:
When a message contains large images they take up a lot of screen space, making it difficult to read the conversation. Especially when there is text between them or when reading on a phone. Also given that the images can't be clicked you have to scroll the image horizontally and make up in your mind what it is about.

- It would be great if the images were limited to a size small enough to quickly read the message and large enough to distinguish the images from each other.
- Click the scaled image to view the full version it in a new tab, let the browser handle the large size.

PR #712, opened 2020-08-24
Rescale large images for readability

fvdm:
_(this is a redo on master branch)_

This CSS fix for issue #612 makes sure that large images are limited to a maximum width of 100% of its parent `<p>` tag.

Because all attached images are already available for download at the bottom of the message, the image does not need to be linked.

I have tested this on the latest stable versions of Safari, Firefox and Edge on macOS and Safari & Firefox on iOS.

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -2224,6 +2224,9 @@ button.conv-checkbox-clear {
     text-align: center;
     top: 75px;
 }*/
+.thread-content	p img {
+    max-width: 100%;
+}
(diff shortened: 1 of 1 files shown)


fvdm:
Screenshots

### Before

Notice how the image from the incoming mail is zoomed to 1:1 scale, requiring a lot of horizontal and vertical scrolling.

freescout-pr711-before

### After

With the fix the image always scales to stay within the viewport.

freescout-pr711-after

PR #753, opened 2020-09-22
Update NL translation

fvdm:


fvdm:
Wow, no good either..

PR #756, opened 2020-09-23
Issue 490: Mailbox Managers and per user management permissions

injektion:
Updates permissions page to edit per user permissions as defined.
Permissions stored in mailbox_user.access and defined in Mailbox ($USER_ACCESS_PERMISSIONS)
Hides pages and options without permission on templates:

- delete button
- mailbox address and name only editable by admin
- settings_menu

Added policy definitions to check permissions on each save page.
Moves email signature to own edit page (similar to Auto-Reply) to keep managers from needing to see breakable settings.


freescout-helpdesk:
It will take some time to check this.

PR #760, opened 2020-09-24
WIP: Add config variable to disable update checks

nebulade:


nebulade:
This currently disables the UI bits to not show the update check or update action button, but also the backend side to not return potential new versions. 

I am not sure if both parts are required, happy to remove the backend code from this PR if this is preferred

nebulade:
Oh that was quick, unfortunately I have put that `WIP:` which is now part of the commit message in master as well. I will keep that at least in mind for the next contribution :-)

freescout-helpdesk:
No problem.

PR #781, opened 2020-10-02
Update de.json

Servery:
Line 201: Missing ":" to display german text with a href correct.
Line 358: Old translation does not sound good in german.
Line 727 - 747: Added some translations


freescout-helpdesk:
Could you please submit pull request into `master` branch.

PR #945, opened 2020-12-17
Enhance and intensify the dashboard card views

nickbe:
Correct the counters to show proper numbers. The counters under "Incoming" and "Mine" as well as "Assigned" are now properly showing not only active messages, but a mixture between ALL messages (Most important with Unassigned) and <active> / <from> for "Mine".

The cards should look like this when all patches are merged:
image

### DIFF
--- a/resources/views/secure/dashboard.blade.php
+++ b/resources/views/secure/dashboard.blade.php
@@ -18,12 +18,50 @@
                         <div class="dash-card-list">
                             @php
                                 $main_folders = $mailbox->getMainFolders();
+                                $count_mine = 0;
                             @endphp
+                            
                             @foreach ($main_folders as $folder)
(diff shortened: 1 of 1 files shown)


fvdm:
The clock icon feels out of style there. What does it indicate?

nickbe:
Hey Franklin, I wanted to indicate how many messages are actually active. In this case 1 active / 14 assigned to me.
I looked through the icons and the clock looked fitting. A little computer screen was what I've looked for. (or a little hammer).

But we could use some seperate icon seperately from the current glyph font.

nickbe:
btw. I'm already working on the same scheme for the detail view.  😀

freescout-helpdesk:
Dashboard now looks good.

PR #946, opened 2020-12-17
Extend layout header to compensate icon position on iOS

nickbe:
The icons were currently not clickable on some devices.


freescout-helpdesk:
Wrong branch. 

Which icons were not clickable?

freescout-helpdesk:
Wrong branch. More info is needed.

PR #959, opened 2020-12-23
fetch emails hook unseen

### LINKED ISSUE #956: Mailbox settings fetch all
It would be helpful if you could configure fetching all / unseen messages in mailbox connection settings. Use case is if mails are fetched from a subfolder where emails are moved into and read status is set.

tillmon:
closes #956


tillmon:
This pull request introduces a hook to overwrite the `unseen` option when fetching emails.

PR #962, opened 2020-12-23
Patch 9

nickbe:


freescout-helpdesk:
There is already `@yield('javascripts')`. It's not enough?

PR #963, opened 2020-12-23
Patch 10

nickbe:


freescout-helpdesk:
It would be good to have all commits related to one task in one pull request.

PR #985, opened 2021-01-08
fetch emails hook days

tillmon:
This introduces a hook to overwrite the `days` option when fetching emails. It is related to #959.


freescout-helpdesk:
Better to do the following:

1) In App/Console/Kernel.php:
```
$schedule->command('freescout:fetch-emails', \Eventy::filter("schedule.fetch_emails.params", [])));
```

Params can be set as: `['--days' => '3', '--unseen' => '1']`

2) Remove `fetch_emails.unseen` hook from FetchEmails.php 

3) Check `$unseen` in FetchEmails.php like this:
```
            if ((int)$unseen) {
                $messages->unseen();
            }
```

freescout-helpdesk:
Closing for no activity.

PR #1220, opened 2021-05-19
Encrypt mail_password

davask:


freescout-helpdesk:
Thanks
1) Pull requests should be sent into `master` branch: https://github.com/freescout-helpdesk/freescout/wiki/Development-Guide#github-workflow
2) Already entered passwords should be converted into encrypted and continue to work.

davask:
Hi,

OK for push on Master.
I m not sure how to handle point 2 that would require a 1 time task, task that I would prefer users to make volontary instead of automatically cause it requires modification of data and can break everything.
Any suggestion?

freescout-helpdesk:
It can be converted without breaking anything.

freescout-helpdesk:
It looks like something went wrong - too many commits.

davask:
Yes, I started over, I m waiting for you to validate the other pull request before pushing that one again

PR #1249, opened 2021-06-04
Customer email copy link

foliovision:
We found we often need to copy the customer email address. This addition of a simple link to copy the email address to clipboard is very handy to us and I believe other users will like it too.
freescout-customer-email-copy-link

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -1202,6 +1202,9 @@ a h4 {
     margin: 0;
     padding: 0;
 }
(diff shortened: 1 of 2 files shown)


freescout-helpdesk:
Behaviour of on click event changed to copying email to clipboard: https://github.com/freescout-helpdesk/freescout/commit/d27b2f9727d424d7c20e8270e4c3808d307380a9

Issue #1355, opened 2021-08-03
PHP 8 / Outdated Dependencies 


admdly:
I've been working on a dev setup for this project, and I've noticed there are a number of issues with PHP 8 and outdated dependencies. 

I understand there is no desire to update to the latest Laravel but are you open to PRs that update dependencies to correct issues, provided these are tested and do not introduce new issues? 

For example, the version of fzaninotto/faker currently required does not work correctly with PHP 8 (implode() error on db:seed) but the latest version appears to work fine without breaking anything.

freescout-helpdesk:
Yes. Feel free to try.

freescout-helpdesk:
https://github.com/freescout-helpdesk/freescout/pull/1357#issuecomment-894013781

PR #1357, opened 2021-08-04
Update dependencies

admdly:
I have updated dependencies where extremely safe to do so as per #1355. To help with review, I have broadly separated my commits into security, php8, and general to help with review. 

From manually reviewing these changes, I don't believe they introduce any issues. However, if there are any issues please tag me and I will resolve them personally.


freescout-helpdesk:
Notice that some vendor files have been overridden: https://github.com/freescout-helpdesk/freescout/blob/dist/composer.json#L112

admdly:
I made no such changes - perhaps this was related to a dependency update? - as far as I can see there have been no significant/breaking changes.

freescout-helpdesk:
What overriding is: https://blog.shahednasser.com/how-to-override-classes-in-php-and-composer/

Some vendor files are overridden: https://github.com/freescout-helpdesk/freescout/tree/dist/overrides
Their corresponding packages' version should be preserved or those files should be manually re-overridden.

admdly:
My mistake, I couldn't see what you were referring to on mobile. Is there any reason for the overrides other than patching outdated versions with fixes? That is, can these be removed assuming the underlying package is now fixed?

freescout-helpdesk:
Unfortunately, not. Updates are complex.

admdly:
I’ll take a look into each of these for the updated packages and fix as appropriate. Some overrides no longer seem needed, but if they are I will ensure the override is updated to apply to the updated package.

admdly:
Working through the overrides, I can see some are simple one line fixes or are fixes to issues that have now been resolved upstream. Is there any reason to keep these given the maintenance burden and blocking of vendor updates? 

Surely it would be better to fix upstream or in the project itself? I will make any changes to achieve this. 

For the more substantial overrides, I will update these for the relevant packages and lock the version appropriately.

freescout-helpdesk:
We've checked updated dependencies, new set of dependencies does not work on a shared hosting:
```
PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 20480 bytes) in /vendor/opis/closure/src/SerializableClosure.php on line 381
PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 20480 bytes) in /vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php on line 149
```

Thanks for your efforts. These works can be stopped.

admdly:
This was still a WIP but sure, I’ll continue my work elsewhere.

PR #1390, opened 2021-08-22
Made it possible to track threads based on a ticket number in the sub…

jensolsson:
…ject.

TicketNumber module required and I also have a small update for this one.
Added .env settings for:
TICKETNUMBER_PREFIX defaulting to "[#"
TICKETNUMBER_SUfFIX defaulting to "] "
TICKETNUMBER_TRUSTSUBJECT defaulting to false (this will toggle this functionality on and off)
This fix will be great for everyone who need to forward tickets to external helpdesks that are not always adhering to standard use of In-Reply-To and Referenced headers.

If the subject matches a conversation_id it will use the first thread in that conversation as its parent messageid

Also have an update of the module TicketNumber that would make the prefix and suffix customizable with the same .env settings as above.

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -309,6 +309,22 @@ class FetchEmails extends Command
                 }
             }
 
+            //Try to get first mesage ID in thread from conversation ID in subject to get some coherence in auto responders that do not send proper headers.
+            if (!$prev_message_id) {
+                $subject = $message->getSubject();
+                $conversation_id = \MailHelper::fetchMessageTicketNumberValue($subject);
+                if (!is_null($conversation_id)) {
+                    $this->line('['.date('Y-m-d H:i:s').'] Using conversation_id ' . $conversation_id . ' matched from subject');
+                    $thread = Thread::where('conversation_id', $conversation_id)->first();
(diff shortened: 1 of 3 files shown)


jensolsson:
I hope I was adhering ok to the project structure. I tried to move all functionality to the TicketNumber module, but it did not seem to be fully possible to do so since I found no way to run code in the fetch email sequence from the module.

freescout-helpdesk:
This can be implemented via custom module and adding a hook to the FetchEmails.php: https://github.com/freescout-helpdesk/freescout/wiki/Development-Guide#actions-and-filters

This functionality may lead to problems and not needed in the core.

mpdude:
@jensolsson Do I get it right that you wrote a module that can parse the conversation ID out of email subjects and use that to assign the mail to the right existing conversation?

jensolsson:
Yes I did, or I rather updated the module that adds conversation Id to the subject so that it would also parse this ID on incoming email. But it was not accepted as a contribution because of low estimated users. @mpdude

mpdude:
Did you publish the module anywhere yourself?

jensolsson:
I did not, and since I modified a module that is payed I don't think I am allowed to do so

PR #1395, opened 2021-08-25
Fix page generation error with multiple agents

ejhayter:
When multiple agents were viewing an email, the following error was seen:

usort(): Returning bool from comparison function is deprecated, return
    an integer less than, equal to, or greater than zero at
    /www/html/app/Http/Controllers/ConversationsController.php:252

Opted for spaceship operator, I was not able to verify that the code is
operating as intended, only that it resolved the generated error.

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -391,13 +391,8 @@ class FetchEmails extends Command
                     return;
                 }
                 $this->line('['.date('Y-m-d H:i:s').'] Message from: User');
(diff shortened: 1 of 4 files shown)


freescout-helpdesk:
Thanks.

1) Commits should be sent to the `master` branch.

2) Here is how it should be fixed:
```
return (int)($a['replying'] < $b['replying']);
```

ejhayter:
Well... my inexperienced with github is showing, it won't forget the target branch, so lets try this again...

PR #1528, opened 2021-11-01
Add support to get module image from modules.json

andrewlimaza:
Add support for module.json 'img' attribute. Use this to set a path to the image for your custom module icon/logo.


freescout-helpdesk:
Does it work? Can you see the image on Modules page?

andrewlimaza:
It works, if the `img` parameter is left out it defaults to the default puzzle icon.

See result:
Screen Shot 2021-11-01 at 15 48 59

PR #1563, opened 2021-11-30
Implement configuration file for default subscriptions (fixes #294)

### LINKED ISSUE #294: FR: Default Notification Settings
Just a Feature that would be nice to have:

I would like to be able to set a default "Notification-Policy" so that as a default every new user gets notifications via Mail for new Tickets.

moqmar:
This lets the user configure the default subscriptions for new users in a separate configuration file (and thus implements #294) instead of manually having to edit `app/Subscription.php`.


freescout-helpdesk:
Please submit to the master branch: https://github.com/freescout-helpdesk/freescout/wiki/Development-Guide#github-workflow

moqmar:
Whoops, sorry. Changed it to master. :+1:

freescout-helpdesk:
This approach is not different from changing app/Subscription.php directly. After updating FreeScout config/subscriptions.php will be overwritten.

Better create a "Default Notification Settings" section in Manage > Settings > General with expandable link "Edit default notifications settings" and store configuration in .env file in DEFAULT_NOTIFICATIONS parameter in json.

moqmar:
You are absolutely right, I completely missed that as I'm using Docker and can just mount it `:ro`. :see_no_evil: 
I'll try to implement the config page in the next days, marking this WIP in the meantime.

moqmar:
Alright, this should work correctly now, please let me know if I'm missing anything. :upside_down_face:

Issue #1150, opened 2021-04-10
Improve accessibility for the visually impaired


neosonic2:
I am a completely blind individual using the VoiceOver screen reading program on macOS to navigate around the Internet, including working inside Freescout.

While the Web app is pretty accessible out of the box, there are many areas of the interface that could use some improvement in terms of how they are read by and made interactible with screen readers.

What plans do the Freescout team have for improving the accessibility of this app for those who use screen readers and other assistive technologies?

I am a Web developer experienced in PHP and the Laravel framework (versions 7 and 8 though, not version 5) as well as Web accessibility standards like ARIA. I would be happy to help contribute to this effort, especially since I already know, based on my usage of Freescout as a blind individual, what elements should be made more accessible and how their accessibility can be improved, without sacrificing their functionality or visual appearance. If you would like me to assist in this endeavor please let me know how I can get started.

Thank you!

freescout-helpdesk:
You can create a simple custom module (Improved Accessibility, for example) and we will add it to https://github.com/freescout-helpdesk/freescout/wiki/Community-Modules. When activated the module will adjust the CSS of the FreeScout. You can create a GitHub repo for the module, so that other can also contribute.

Here is the info on developing modules: https://github.com/freescout-helpdesk/freescout/wiki/Modules-Development

neosonic2:
Thanks for your reply. However, the accessibility changes/improvements that are needed would require modifying the HTML of the core Freescout application, i.e. to add or improve ARIA roles and other attributes. Could this be done by creating a custom module as you suggest? I'm not going to be changing anything visually so I probably wouldn't be editing any CSS; I would mostly just be making HTML changes.

freescout-helpdesk:
Yes. Just use Actions and Filters.

neosonic2:
After some further investigation, it looks like a lot of the accessibility changes I need to make can be done to the HTML within the resources/views folder. For example, there are unlabeled buttons that change the conversation type from e-mail to phone and are located in resources/views/conversations/create.blade.php that could be edited. Once these changes are made, can I create a pull request here to merge my accessibility enhancements into the core project?

freescout-helpdesk:
Yes.

fulldecent:
@neosonic2 Could you please provide one example of a thing we can change. Let's start with that and then make a process to fix more

neosonic2:
[...]

PR #1578, opened 2021-12-08
Improve accessibility of modal dialogs (#1150)

philippkraft:
Change aria-hidden attribute to aria-modal attribute to make modal dialogs accessible for screen readers

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -2112,11 +2112,11 @@ function triggerModal(a, params)
     }
 
     var html = [
-    '<div class="modal '+(params.no_fade == 'true' ? '' : 'fade')+'" tabindex="-1" role="dialog" aria-labelledby="jsmodal-label" aria-hidden="true">',
+    '<div class="modal '+(params.no_fade == 'true' ? '' : 'fade')+'" tabindex="-1" role="dialog" aria-labelledby="jsmodal-label" aria-modal="true">',
         '<div class="modal-dialog '+modal_class+'">',
             '<div class="modal-content">',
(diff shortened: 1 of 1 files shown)


philippkraft:
This pull requestis tested with JAWS on Windows. @neosonic2 will test this PR for two other major screen readers.

neosonic2:
Testing this PR with other popular screen reading programs is taking slightly longer than anticipated as I am having problems with my local macOS development environment. I will comment again when I have that sorted out and have tested the PR with VoiceOver on macOS and NVDA on Windows, the other two most popular screen reading packages available besides JAWS.

neosonic2:
I have finally been able to test out this PR using several other popular screen readers and below are my results. I tested this PR against the production version of Freescout used at my workplace, which is version 1.7.20. In both tests, I opened a conversation and clicked the Merge button in the More Actions menu to open that modal dialog.

When testing VoiceOver on macOS 12 Monterey against the production version of Freescout, it was able to recognize that the Merge Conversations dialog had opened, but the only thing it read out in the dialog was the Close button. No other elements were present.

However, this PR performed much better; VoiceOver was able to read everything in the Merge Conversations dialog without any issues.

NVDA 2021.3 running on Windows 11 produced similar results, except that with the production version of Freescout, NVDA couldn't see anything inside the dialog, not even the Close button. It detected that the Merge Conversations dialog had opened but reported that it contained no elements. Testing against this PR though yielded a dialog that was fully readable.

So I'd say this pull request accomplishes its stated goal of improving the accessibility of modal dialogs in Freescout across all major screen reading applications - JAWS for Windows, VoiceOver for macOS, and NVDA for Windows. I vote that it be merged into the upstream Freescout distribution.

PR #1593, opened 2021-12-17
Added hooks for custom fetch email providers

Wouter0100:
This allows us to create custom importers for e-mail. It will use the FetchEmails class to add the e-mails by "simulating" the Message and Attachment classes. An example for Postal, [here](https://github.com/Wouter0100/freescout-postal/).

It would also be possible to use this for example for AWS SES Incoming emails and other e-mail providers providing incoming e-mail abilities.

No behaviour changed by default.

https://user-images.githubusercontent.com/864520/146613993-d801d2d6-bb89-4205-912e-f6cc17c4e205.mp4

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -75,8 +75,8 @@ class FetchEmails extends Command
 
         $this->extra_import = [];
 
-        // Get active mailboxes
-        $this->mailboxes = Mailbox::get();
+        // Get active mailboxes with the default in_protocols
+        $this->mailboxes = Mailbox::whereIn('in_protocol', array_keys(Mailbox::$in_protocols))->get();
 
         foreach ($this->mailboxes as $mailbox) {
             if (!$mailbox->isInActive()) {
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -75,8 +75,12 @@ class FetchEmails extends Command
 
         $this->extra_import = [];
 
-        // Get active mailboxes with the default in_protocols
-        $this->mailboxes = Mailbox::whereIn('in_protocol', array_keys(Mailbox::$in_protocols))->get();
+        if (Mailbox::getInProtocols() === Mailbox::$in_protocols) {
+            $this->mailboxes = Mailbox::get();
+        } else {
+            // Get active mailboxes with the default in_protocols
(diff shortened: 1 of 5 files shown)


Wouter0100:
Thinking about it, this still needs some work for the default FetchEmails command to skip the non-default protocols, so please don't merge yet as I will attempt to add that as soon as possible.

Regardless, I'm still thinking about your opinion on this and like to know if this is something that would be able to get merged at some point.

Wouter0100:
I haven't had time for the above. I'm still planning on doing that ASAP, as we need it ourselves as well :).

Regardless, as you've most likely seen as well - ideally we would move the handling of e-mails in the FetchEmails to a more accessible location for other scripts/modules to be able to process e-mails. But, without it the proposed solution should work just fine.

Wouter0100:
Implemented a fix for that as well, see [here](https://github.com/freescout-helpdesk/freescout/pull/1593/files#diff-e12b566d766fcc92a657c5895fa11b550afb364c47c64512ee5bbe9a5f639666R78-R79).

Wouter0100:
We deployed in production now. Got a mix of Mailboxes working perfectly fine.

freescout-helpdesk:
Some changes are needed:

1) Replace `@foreach($mailbox->getInProtocols() as $key => $name)` with @action()

2) `div` with default settings must be visible by default without need to show it via JS.
 
3) This `$this->mailboxes = Mailbox::whereIn('in_protocol', array_keys(Mailbox::$in_protocols))->get();` should be executed only if there are extra protocols in `Mailbox::getInProtocolDisplayNames()`;

4) `$response = \Eventy::filter('mailbox.fetch_test', $response, $mailbox);` - this filter should set $response['tested'], if this flag is set default checks should be skipped.

Wouter0100:
I have added 2, 3 and 4. I however don't really understand what you mean with 1. As far as I know, @action is not a replacement for a foreach statement. And besides that, @action is not returning data?

Thanks for the feedback though.

freescout-helpdesk:
See for example this: https://github.com/freescout-helpdesk/freescout/blob/c4a4f33633ef06ec8c9fc5cb668bb75cce2c7793/resources/views/conversations/view.blade.php#L152

Wouter0100:
Hmm, I see. In that case you want us to manually do that logic? 

I've changed it. Could you please verify this is as you suggested?

freescout-helpdesk:
If you wish feel free to share your custom module with others: https://github.com/freescout-helpdesk/freescout/wiki/Community-Modules

PR #1622, opened 2022-01-05
"convsersation" typo

TylerB260:
Hey there - apologies if I'm doing anything wrong since this is the first time I've done a pull request.

I noticed this typo in my notification history this morning and it would appear that it's been there for quite some time. I fixed it in the thread.php script, and all of the applicable translation files. However I'm thinking that updating the actual string in the script is going to break old notifications, right? If there's anything in place for situations like this then disregard. :) 

Thanks!


TylerB260:
Gotcha - my bad. I’ll close this and make a new one since I started on the wrong branch.

PR #1651, opened 2022-01-20
Added Conversations sorting by subject, Number, Creation Date in Mailbox view page

alinmbb:
The sorting can be done by clicking on the column names while showing correct sorting arrow next to the name when necessary. 

By default it it's using the created_at ascending order sorting.


freescout-helpdesk:
Thanks. We'll check it.

PR #1745, opened 2022-02-18
Alternative signatures

Lars-:
One of the things I really miss in Freescout is the ability to change the signature per conversation. Most of my conversations are Dutch, but when I need to send an English mail, my only option is to send the mail with the Dutch signature. 

In this pull request I added alternative signatures per mailbox. 

The alternatives can be saved per mailbox (/mailbox/settings/{id} route) and can also be changed and deleted there. They are then stored in a custom table so the mailbox rows won't get too large.

When you start a new conversation (or reply to one), the signature field is added left to the status field if there actually are alternatives for that mailbox. If there are no alternative signatures, the field is not displayed.
The selected signature is saved with the conversation so that you don't have to select it again for that conversation.

Of course, everything is translatable.


freescout-helpdesk:
Great. But better to implement it as a custom module: https://github.com/freescout-helpdesk/freescout/wiki/Modules-Development
You already have all the code. Just add needed hooks to the core of FreeScout.

freescout-helpdesk:
Please submit just commits with hooks.

PR #1750, opened 2022-02-19
Add RTL and Fa-IR language support for project.

rhashnull:
As you know Persian (Fa-IR) language is Right-to-left written language.
So if we want to have the project supporting the Fa-IR lang, we have to add RTL support to the project style.
Here I share the result with you guys. let me know if there is any issue of bug in this pull request.


freescout-helpdesk:
Few updates are needed:

1) Remove .envrc file

2) Add to config.php a list of RTL languages:
`'locales_rtl'         => ['fa'],`

3) Add to /app/Misc/Helper.php the `isLocaleRtl()` method to check if current language is RTL using the config parameter.

4) If some language is RTL add `dir="rtl"` attribute to `<html>` tag and "rtl" class to the `<body>`

5) Rename `rtl.css` into `style-rtl.css`.

6) There are other languages which may need their own font. Add `locale-{{ app()->getLocale() }}` to the body and load Yekan font only for `.locale-fa`.

7) By some reason your editor removes 4-space indents and breaks lines which makes it difficult to check what has actually changed in blade files. Need to be fixed, so that we can see what actually has changed: https://github.com/freescout-helpdesk/freescout/pull/1750/commits/cc8e5da17b3c0945f680d564b996167e450dd1d5#diff-66a3ba2334a6e1eebc24807619f1d4e991f1e8f58c5b5af2c787198d62c547d7

rhashnull:
Thanks for the review, I'll figure them out.

freescout-helpdesk:
We'll check it after some time

freescout-helpdesk:
composer.lock need to be removed from the pull request.

rhashnull:
Hi there, I reviewed your comments and done them one by one. can you please check them again before any merge conflict occurs again.

PR #1806, opened 2022-03-09
add check for null passwords

zschroeder6212:
Add code to handle null passwords in the safePassword function.

This was causing errors when trying to access the '/app-settings/emails' page for the first time.


freescout-helpdesk:
Thanks. We've fixed it like this: https://github.com/freescout-helpdesk/freescout/commit/2e71180f7ce748f102d879134afac3a0f14a222f

PR #1865, opened 2022-03-31
Add header to enable easier mail filtering

DavidAnderson684:
By adding a header that references the mailbox name, it is easier for support agents to set their desired filter rules. For example, an individual support agent may want notifications from some mailboxes, but not others. He turns on notifications on his account, but then has to use email filters to delete the ones he did not want.

Keep in mind that pull requests should be sent to the `master` branch! See https://github.com/freescout-helpdesk/freescout/wiki/Development-Guide#github-workflow

Now you can delete this text and type the description of your pull request...


DavidAnderson684:
Wrong branch.

PR #1866, opened 2022-03-31
Add header indicating mailbox name

DavidAnderson684:
By adding a header that references the mailbox name, it is easier for support agents to set their desired filter rules. 

For example, an individual support agent may want notifications from some mailboxes, but not others. He turns on notifications on his account, but then has to use email filters to delete the ones he did not want. Or, he wishes to redirect notifications for different mailboxes to different final destinations.


freescout-helpdesk:
No need to expose mailbox name in headers. Feel free to implement as a custom module: https://github.com/freescout-helpdesk/freescout/wiki/Modules-Development

DavidAnderson684:
@freescout-helpdesk My post explains the need.

PR #1882, opened 2022-04-05
Add filter allowing customisation of outgoing headers

DavidAnderson684:
I wish to add a header to outgoing mail (see: https://github.com/freescout-helpdesk/freescout/pull/1866). I was told to do it as an external module.

This PR adds the necessary filter to allow that.


freescout-helpdesk:
Just notice that filter name was changed to `jobs.send_reply_to_customer.headers`

PR #2034, opened 2022-06-08
Enhance of customer form

nykopol:
Two commits to enhance the customer form.

The first is to use navigator build-in helper for emails, phone and url fields.
The second is to edit and show the phone type of phone numbers (home, mobile, fax, ...). This informations was already in the database but was not usable.


freescout-helpdesk:
Looks good. Just translations are not stored in such `phone_type.work` form in FreeScout. See how other translatable strings are stored.

nykopol:
Ok, I did the modification about translation keys.

Issue #2141, opened 2022-08-05
Conversation modals for kanban module need work


vtq221:
The modals in the kanban module break the UI for all kinds of actions that you'd expect to be able to do.

The x_embed URL param does not persist or is not always recognised, resulting in broken routes and the main menu appearing twice.

Behaviour that you would want is for the UI not to break, and for all visible UI routes to work.

I think some of the UI elements visible in the modal are unecessary.

I also don't think the conversation should redirect to the conversation list when in kanban - the kanban itself serves that purpose.

I've fixed some of these issues already in my fork, so happy to PR once done if maintainers agree with these suggestions.

freescout-helpdesk:
Feel free to submit a PR - we'll take a look.

virtiqa:
OK will do once I've done some more testing.

Are you comfortable using javascript APIs that won't work with IE11?

This in particular: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams 

Support is here: https://caniuse.com/urlsearchparams

freescout-helpdesk:
main.js has a `getQueryParam()` function for retrieving get parameters.

vtq221:
Closing - changes have been merged

PR #2143, opened 2022-08-08
Fixes to conversation view in kanban modals

vtq221:
As discussed in #2141, changes to preserve the x_embed param while conversations are worked on in an iframe (e.g. with the kanban module).

Fixes some routes in views, and checks for x_embed and appends to redirect URL.

Also keeps conversation on same page in iframe and does not redirect to conversation list (as conversation list is available in the kanban master window).

Note that the built in getQueryParam() function in main.js didn't quite do what I needed to, so this PR is still introducing SearchParams API: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams 

If this is a problem I can write an appendQueryParam() function, just let me know.

I also have code for hiding various components in the x_embed window using JS/CSS, but I've left it out because it might be more of a user preference than a core requirement. Can easily be added with a module etc.

### DIFF
--- a/app/Http/Controllers/ConversationsController.php
+++ b/app/Http/Controllers/ConversationsController.php
@@ -390,7 +390,7 @@ class ConversationsController extends Controller
                     // Determine redirect
                     // Must be done before updating current conversation's status or assignee.
                     $redirect_same_page = false;
-                    if ($new_user_id == $user->id) {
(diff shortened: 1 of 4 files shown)


vtq221:
This PR got messy for some reason - will do fresh PR from a rebase based on recent code so the changes are clearer, and will remove the dependency that would break IE11

Issue #2167, opened 2022-08-19
New action for conversation view


ilyakonovalenko:
Hello,

Please add a new action to show content in modules between customer information and previous customer conversations.

File `resources/views/conversations/view.blade.php`, before line 280:
```
@action('conversation.before_prev_convs', $customer, $conversation, $mailbox)
```

In the case we can display some information about customer in full width.  There is an action `customer.profile_data`, but its width is not 100% of block content, less than previous conversations block width.

freescout-helpdesk:
Feel free to submit a pull request.

PR #2169, opened 2022-08-19
New action before_prev_convs per issue #2167 + An ability to keep current status and assignee per issue #2180

ilyakonovalenko:
Additional action per issue #2167


ilyakonovalenko:
Added an ability to keep current conversation status and assignee per mailbox, per issue #2180.
The changes have been tested on my env and should be fully compatible with any setups do not want the ability.

freescout-helpdesk:
In the future better not to mix different tasks into one pull request.

PR #2224, opened 2022-09-12
Make conversation history table wider in customer profile, Fix #61

### LINKED ISSUE #61: Make conversation history table wider in customer profile
todo

fulldecent:
Before

Screen Shot 2022-09-12 at 12 05 24

After

Screen Shot 2022-09-12 at 12 05 26

---

Requesting testing by others to confirm this does not mess up other uses of conversation view, specifically with paging.


freescout-helpdesk:
Good job. We've adjusted it a bit.

PR #2244, opened 2022-09-15
Use n-dash for date range

fulldecent:
N-dash is typically used for ranges of numbers rather than a minus symbol.


freescout-helpdesk:
Thanks. But we like it more as it is now.

PR #2366, opened 2022-11-04
french translation corrections

e-rwan:
just a few french corrections.
**"Type": "Tapez"** -> "tapez" means typing in french, the word for "type" as in "kind, sort" in frenchr is ... "type"


e-rwan:
I also have corrections for some modules,.
workflow in particular have several translation that dosen't make sense/are out of context. I had to check the json file to understand the meaning.
but I wouldn't know where to post thoose?

freescout-helpdesk:
See https://freescout.net/modules-faq/#translate-modules

e-rwan:
every module translation is empty(whatever the language) even thought the tranlation exist.
Capture
is that normal behavior?

PR #2397, opened 2022-11-18
[proposition] inner scroll with toolbar and menu always visible

e-rwan:
hello,
I found it a bit frustrating to have to keep scrolling up and down to access menus and toolbars.
so I made this which IMHO makes navigation more agreeable, especially on big screen.

some legacy browsers may not support calc() or sticky, that shouldn't cause any issue here.
<br><br>

the footer is problematic with this function as sometimes it's here, sometimes not and darkmode disables it entierly.
there are several solution for that but it would require to change some of the core css.

```
    .footer {
        display:none;
    }
```
<br>
those div sometimes have a min-height declared in the html element. I couldn't find out why?

```
    .content-2col, #conv-layout, #conv-layout-main{
        min-height: unset !important;
    }
```
<br><br>
freescout_inner-scroll

<br><br>
tell me if this is of any interset to you, otherwise I'll just make a module of it.

erwan


freescout-helpdesk:
Thanks, but we'll leave it as is. Feel free to share as custom module.

PR #2434, opened 2022-12-01
Search improvments

calebbaker194:
This will allow searching by customer first and last name to find threads or customers. It also ignores cases on more things.


freescout-helpdesk:
In MySQL "like" is case insensitive by default. It does not support "ilike". So "ilike" changed back to "like".

calebbaker194:
That makes sense. I set mine up with a PostgreSQL instance.

PR #2561, opened 2023-01-17
Fixed the translation of the status "Pending"

il2020:
"Pending" translates to "В ожидании" but not "Отвечен".
https://translate.google.com/?sl=en&tl=ru&text=Pending&op=translate


ilyakonovalenko:
It depends on the use case. Basically it means that ticket is answered and is pending customer reply, so "Отвечен" is also correct.

il2020:
> It depends on the use case. Basically it means that the ticket is answered and is pending customer reply, so "Отвечен" is also correct. "В ожидании" also means "Waiting" which is a bit unclear.

In my opinion, "Отвечен" looks like the final status as "Closed", that is, the processing of the ticket is complete, but it is not.

freescout-helpdesk:
We'll leave it as is.

PR #2744, opened 2023-02-19
Add "Followers" folder

avenjamin:
Add "Followers" folder to show followed conversations.

### DIFF
--- a/app/Conversation.php
+++ b/app/Conversation.php
@@ -195,6 +195,13 @@ class Conversation extends Model
     public static $starred_conversation_ids = [];
 
     /**
+     * Cache of the conversations followed by user.
+     *
+     * @var array
+     */
+    public static $followed_conversation_ids = [];
+
+    /**
      * Cache of the app.custom_number option.
      */
     public static $custom_number_cache = null;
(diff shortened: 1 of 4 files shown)


freescout-helpdesk:
Thanks, but not needed in the core. Can be added via a custom module.

avenjamin:
I disagree.

In the Core you have the ability to star a conversation and there is a Starred folder.
You have the ability to follow and unfollow a conversation so why no Following folder?

I get that you have a Followers Module already, but that only adds additional functionality to add other followers.

Currently to see conversations you're following you have to go to the Search menu in the top right, then click "Conversations I'm following" which takes you to a different layout, then to go back to the folders you have to use the back button in the browser.

This makes the Core Follow/Unfollow feature more accessible and keeps you in the same layout/view as the other folders.

Please reconsider.

Thanks

freescout-helpdesk:
https://github.com/freescout-helpdesk/freescout/wiki/FreeScout-Development-Process

avenjamin:
Would you consider adding this to the Followers module?

I don't currently have a license for it so I can't contribute directly, but if you really don't think this belongs in core then the Followers module would make the most sense.

PR #2982, opened 2023-05-05
Update folders counters in background

Lars-:
When you have a lot of mailboxes, a lot of users and you use tags and the custom folders plugin excessively, every action becomes extremely slow (on serious dedicated hardware, sending a mail took 20s, deleting the same), because each action triggers a updateCounters for each folder. 

In this PR, the updateCounters function is moved to a job. When you have an extreme amount of folders, it will take a few seconds before the counters update, but everything else is way faster. When you have 1 mailbox with some folders, you won't notice a difference.

I also built a small part that checks if the job is already queued. If it is, it won't be added again. So removing 10 mails won't add all folders 10 times to the queue. Feel free to make this optional via the settings or something (you could use dispatchNow instead of dispatch when some setting is set for example)

### DIFF
--- a/app/Folder.php
+++ b/app/Folder.php
@@ -223,44 +223,11 @@ class Folder extends Model
 
     public function updateCounters()
     {
-        if (\Eventy::filter('folder.update_counters', false, $this)) {
+        if(\Cache::has('updating_folder_' . $this->id)) {
             return;
         }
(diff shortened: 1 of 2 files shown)


freescout-helpdesk:
The following .env parameter can be used to enable updating folder counters in the background:
```
APP_UPDATE_FOLDER_COUNTERS_IN_BACKGROUND=true
```

Issue #2901, opened 2023-04-06
Upgrade via tools.php script


AdyGCode:
PHP: 8.1.x
Freescout: 1.8.69

When attempting to use the tools.php script to upgrade (due to open_basedir restrictions) the script reports PHP 7.x is required.
Could script be updated to allow for PHP 7.4 or later, so as to allow the upgrade with less issues?

This also takes into account that PHP7 is (close to) End Of Life.

![2023-04-06 14_43_44-FreeScout Tools and 9 more pages - Work - Microsoft​ Edge](https://user-images.githubusercontent.com/60425995/230293831-2a8087c3-7063-4f70-8ba4-2b7e3afb9a8e.png)

freescout-helpdesk:
Fixed in the `master` branch.

PR #3063, opened 2023-05-31
fix: tools.php - PHP Version comparison error fix

AdyGCode:
PHP Version check disallows all PHP versions unless before 7.1 due to >=

Correction to allow all PHP versions from 7.1 and after.

[Related to issue #2901 ](https://github.com/freescout-helpdesk/freescout/issues/2901)


freescout-helpdesk:
The condition is correct. Check your console PHP version.

PR #3131, opened 2023-06-23
Fix Japanese Text Garbled. 

totake0224:
Fix Japanese text in attachment filename, subject ,sender, body garbled or display '?' .
this commit is for APP_NEW_FETCHING_LIBRARY=false
after this change there is no garbled text in my mailbox yet.

I will be very happy. if this request accepted. and I could contribute freescout.
this is my first pull request. I have read development guide and I followed it.
if anything strange. please tell me.


totake0224:
I notice I'm using dist branch.
I will redo using master branch.

PR #3263, opened 2023-08-14
Speed up eventy event firing

OttoWinter:
Freescout appears to use `Eventy` to allow modules to customize certain aspects of freescout, for example the user `getFullName`:

https://github.com/freescout-helpdesk/freescout/blob/40fc84546168c7fed1eacf55759ff008371815f4/app/User.php#L216

Unfortunately, this eventy library appears to be quite inefficient, especially when you have large amounts of listeners (it does a `O(n)` search through _all_ listeners globally to find matching ones, plus each time does a `O(n log n)` sort by priority).

For example, on an instance I was profiling, there were ~120 listeners, thus making a single (!) `User::getFullName` take a whopping 3ms (with ~100 `getFullName` calls for listing assignable users this has a major impact; and that's just for the `getFullName` calls).

This PR overrides Eventy, to replace its internal data structure to be more efficient (keyed by the hook).

In total, this reduced mailbox loading times for my organization from ~1.5s to ~0.5s.

Note: I barely know PHP (in fact most of this code was generated by Copilot), so the code probably is very bad ^^ If there's something that can be done better let me know.

Also note that technically I'm changing the (public) `getListeners` function argument types here. This method only appears to be used inside Eventy, but this could still cause problems (it works on my instance though).

### DIFF
--- a/overrides/tormjens/eventy/src/Action.php
+++ b/overrides/tormjens/eventy/src/Action.php
@@ -14,18 +14,16 @@ class Action extends Event
      */
     public function fire($action, $args)
     {
-        if ($this->getListeners()) {
-            $this->getListeners()->where('hook', $action)->each(function ($listener) use ($action, $args) {
-                $parameters = [];
-                for ($i = 0; $i < $listener['arguments']; $i++) {
-                    if (isset($args[$i])) {
-                        $parameters[] = $args[$i];
-                    } else {
-                        $parameters[] = null;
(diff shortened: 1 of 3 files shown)


freescout-helpdesk:
Thanks. Very good improvement.

raramuridesign:
@OttoWinter Thank you for this, it has improved our installations 
M.

PR #3296, opened 2023-08-24
make mail receive time configurable.

totake0224:
this change will make configurable that,
 option 1: Mail fetch time as receive time (current, default)
 option 2: Mail server receive time as receive time (added)
under option 2  user could know actual mail receive time after system down or fetching trouble.

setting data will be saved in .env file


freescout-helpdesk:
Also keep in mind that Date in the email headers has time zone, so it will need to be taken into account.

Issue #3196, opened 2023-07-18
imap_utf8 could not decode iso-2022-jp-ms properly.


totake0224:
version 1.8.86 
APP_NEW_FETCHING_LIBRARY=false

I tested latest freescout. and I found mail subject garbled. 
I tried to identify which decode function make garbled. and I found it imap_utf8.
garbled image(attachment filename decoded correctly.)
![Screenshot from 2023-07-18 19-55-00](https://github.com/freescout-helpdesk/freescout/assets/42058913/0c58923a-fd49-4551-98e4-55652e44d20c)

correct image
![Screenshot from 2023-07-14 19-44-00](https://github.com/freescout-helpdesk/freescout/assets/42058913/c7af9a2d-9ada-4dd6-b03d-469292217d7b)

This is how I identify the function.
I added \Log function before and after imap_utf8 and I got this.
Modified code
> app\Misc\Mail.php  around line number 955

> \Log::error("imap_utf82:" . $subject);
            $subject_decoded = \imap_utf8($subject);
            \Log::error("imap_utf82_decoded:" . $subject_decoded);

Log output
> 
Level | Context | Date | Content
-- | -- | -- | --
error | production | 2023-07-18 20:06:55 | imap_utf82_decoded:(株)あいうえお�かきくけこ
error | production | 2023-07-18 20:06:55 | imap_utf82:=?iso-2022-jp-ms?B?GyRCLWokIiQkJCYkKCQqfGIkKyQtJC8kMSQzGyhC?=

totake0224:
I also confirm mb_decode_mimeheader decode attachment filename properly by Log output.

totake0224:
I'm very sorry. decodeSubject function is very complex. it need time for me modifying it without side effect.

freescout-helpdesk:
We need just 3 things and nothing else:

1. How it looks when garbled.
2. How it should look.
3. How it looks in the original code of the email (or attach .EML file).

Number 3 is missing.

See how to report decoding issue: https://github.com/freescout-helpdesk/freescout/issues/3177

totake0224:
I have confirmed this fixed.
Thank you.

PR #3310, opened 2023-08-29
Fix Japanese subject garbled.

totake0224:
I confirmed iso-2022-jp-ms chars is not decoded properly. it decoded properly only program reach to mb_decode_mimeheader at last resort.
otherwise, it's decoded inproperly (in many cases partially decoded).
I added only iso-2022-jp-ms coding use mb_decode_mimeheader.
I believe this fix could resolve the garble and make no side effect.


freescout-helpdesk:
Please follow this instruction https://github.com/freescout-helpdesk/freescout/issues/3196#issuecomment-1640078379 and we'll try to fix it.

PR #3320, opened 2023-08-30
Add S3 support

veebkolm:
Add native S3 support by adding configuration options, Laravel's native S3 dependencies and modifying Attachment class to support those changes


freescout-helpdesk:
Pull requests should be sent to `master` branch: https://github.com/freescout-helpdesk/freescout/wiki/Development-Guide#github-workflow

freescout-helpdesk:
Another thing - this feature adds 37 MB of `aws-sdk-php` code, we don't need it in the core of the app.

99% of FreeScout users will never use this feature, so better to implement it via a custom module: https://github.com/freescout-helpdesk/freescout/wiki/Modules-Development

PR #3341, opened 2023-09-05
Patch 2

Pablohn26:
Improve Spanish translation. 
* Adding `¿` and `¡` marks
* Fixing some words without accent mark


Pablohn26:
This PR is wrong because I choose a worng branch. Sorry. Redoing soon.

PR #3399, opened 2023-09-15
Support CI

fulldecent:
This will make continuous integration easier


freescout-helpdesk:
Or maybe better to use `--yes` flag like in https://github.com/freescout-helpdesk/freescout/blob/dist/tools/update.sh

freescout-helpdesk:
FYI. We'll close this. We'll organize application installation in a bit different way in the `test.yml`

PR #3408, opened 2023-09-19
Add filter `fetch_emails.mailbox_to_save_message`

baldugus:
I found it useful to change the destination mailbox to the message based on the folder it was retrieved.


baldugus:
Changed the variable name to snake case, as the rest of the code does.

PR #3752, opened 2024-02-01
Fix for the Italian translation.

PeopleInside:
Fixing Italian translation.
Email subject is not "Soggetto email" but should be translated as "Oggetto email".

This Pull Request fix that.


freescout-helpdesk:
It looks like something got messed up with commits in this PR.

Translations can also be sent this way: https://github.com/freescout-helpdesk/freescout/wiki/Translate

PR #3791, opened 2024-02-15
HTML messages not displayed correctly

smilemittal:
HTML messages not displayed correctly

### DIFF
--- a/app/Thread.php
+++ b/app/Thread.php
@@ -332,7 +332,7 @@ class Thread extends Model
             $body = $this->body;
         }
 
-        $body = \Helper::linkify($this->getCleanBody($body));
+        //$body = \Helper::linkify($this->getCleanBody($body));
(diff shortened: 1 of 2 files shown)


app/Thread.php
```diff
@@ -332,7 +332,7 @@ public function getBodyWithFormatedLinks(string $body = '') :string
             $body = $this->body;
         }
 
-        $body = \Helper::linkify($this->getCleanBody($body));
+        //$body = \Helper::linkify($this->getCleanBody($body));
```
mokkin, line 335: The desired link is displayed in the message after uncommenting this line @smilemittal
Therefore a few other messages are displayed incorrectly and shrinked to half screen size:

grafik

smilemittal:
Ok @freescout-helpdesk Let me do that

PR #3865, opened 2024-03-11
fix issue for HTML purifier

smilemittal:
I have ignored "mso" comments in HTML, So it did not hide links or any other content in the commented block.
Also, I have whitelist style for HTML, So HTML content render properly.

### DIFF
--- a/app/Thread.php
+++ b/app/Thread.php
@@ -315,6 +315,12 @@ class Thread extends Model
         // Cut out "collapse" class as it hides elements.
         $body = preg_replace("/(<[^<>\r\n]+class=([\"'][^\"']* |[\"']))(collapse|hidden)([\"' ])/", '$1$4', $body) ?: $body;
 
+        // Remove only the <!--[if !mso]><!--> and <!--<![endif]--> around the <a> tag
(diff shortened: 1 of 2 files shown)


freescout-helpdesk:
It would be good to get some description. Some of these changes we can't accept.

smilemittal:
I have update the description above

freescout-helpdesk:
Can you provide examples.

smilemittal:
`<a href="https://www.maklerinfo.biz/maklerportal/?show=kunde&kunde=5002970262&Job_archiveintrag=b50365561a57c1fc97fc">to the document</a><! --<![endif]-->
        `
        
In the code above, a link button goes missing because it gets removed during the filtering process.

freescout-helpdesk:
> I have ignored "mso" comments in HTML, So it did not hide links or any other content in the commented block.

https://github.com/freescout-helpdesk/freescout/commit/0826102f6e97d918dd5091fd4a0bbf4679870e76

Your purifier config still does not enable `<v:roundrect>` elements. And we don't see any solutions to allow it for now.

freescout-helpdesk:
> with the purifier I just enabled style tag and CSS class

Unfortunately it's no use as HTML Purifier library removed `<v:roundrect>` element.

PR #3932, opened 2024-04-01
Allow limiting command freescout:fetch-emails to specific mailboxes

ohader:
This change introduces an optional command option `mailboxes`, which can be used to limit fetching messages to particular mailbox ids.

The following example fetches all messages of the last 500 days for the specific mailboxes 5 and 7 only. By omitting the new `mailboxes` option, any message of any mailbox would have been fetched.

```
php artisan freescout:fetch-emails --mailboxes=5,7 --unseen=0 --days=500
```

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -31,7 +31,7 @@ class FetchEmails extends Command
      *
      * @var string
      */
-    protected $signature = 'freescout:fetch-emails {--days=3} {--unseen=1} {--identifier=dummy}';
+    protected $signature = 'freescout:fetch-emails {--days=3} {--unseen=1} {--identifier=dummy} {--mailboxes=0}';
 
     /**
      * The console command description.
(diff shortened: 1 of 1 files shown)


freescout-helpdesk:
Can you please rename this parameter into `mailboxes` and allow to specify multiple mailboxes using comma: `--mailboxes=7,1,5`

ohader:
The Laravel ways seems to be `{--mailbox=*}` in the signature, which would lead to `php artisan freescout:fetch-emails --mailbox=7 --mailbox=1 --mailbox=5 --unseen=0 --days=500` for fetching multiple mailboxes.

See https://laravel.com/docs/11.x/artisan#option-arrays

In case you prefer a string `7,1,5` instead, I could adjust it like that as well (with string parsing & sanitization).

freescout-helpdesk:
See https://laravel.com/docs/5.5/artisan

ohader:
I've adjusted the commit and rebased it on top of the `master` tip.

freescout-helpdesk:
> In case you prefer a string 7,1,5 instead, I could adjust it like that as well (with string parsing & sanitization).

Yes, can you do it this way.

ohader:
> > In case you prefer a string 7,1,5 instead, I could adjust it like that as well (with string parsing & sanitization).
> 
> Yes, can you do it this way.

Done. I've adjusted it to be `--mailboxes=1,2,3`.

Issue #3880, opened 2024-03-15
Browser notification not working for more than one user


yoseabe:
Hello, 
when testing freescout as an admin the browser notification works fine, but as soon as I one more user to have access to the mailbox the browser notification stops working for me(admin) and for all other users, which means only one user is able to recive browser notifications.
Is there a solution for this issue?
We want to be able to use one Mailbox with multiple users that work on it.

PHP version: 8.3
FreeScout version: latest 
Database: MySQL
Are you using CloudFlare: No

freescout-helpdesk:
We can't reproduce the issue. Are you using any non-official modules?

yoseabe:
Yes, everything is OK and I am not using any modules

freescout-helpdesk:
If you'll manage to find the reason feel free to submit a pull request.

yoseabe:
@freescout-helpdesk am I the only one getting this issue? I installed by using Interactive installation bash-script (Ubuntu / Debian / etc). I tried changing php Versions 7.4 and 8.0 both of them with the same issue.

freescout-helpdesk:
Fixed in the `master` branch and will be published in the next release.

PR #3985, opened 2024-04-19
Update Subscription.php

LC-Ilari:
Fixed a bug where only the last created user who enabled the setting received the web notifications. 

Issue was described in https://github.com/freescout-helpdesk/freescout/issues/3880.


freescout-helpdesk:
Thanks. We've pushed a bit adjusted solution: https://github.com/freescout-helpdesk/freescout/commit/29d8c41395b83ad023d9db7d610588fecbff4a4c

PR #4221, opened 2024-09-09
Remove only the <![if !supportLists]> or <![if !vml]> and <![endif]> around the elements

il2020:


freescout-help:
`<![if !supportLists]>` - this is not correct HTML comment. So we don't plan to include this into the app. You can implement custom filtering of the email body using a custom module.

PR #4267, opened 2024-10-04
add php parameter to update.sh

jkmssoft:
The update.sh is not working on ionos, as php is very old version, we need to call php8.3.
With this filechange it is now possible to use --php php8.3 as parameter to set the path/bin name of php to use.

But the update destroys our changes in the artisan and .htaccess files.
Any idea to solve this?
Maybe with a post_update.sh script?

### DIFF
--- a/tools/update.sh
+++ b/tools/update.sh
@@ -6,15 +6,38 @@
 # 
 # Available flags:
 # --yes - answer yes to everything
+# --php - set path to php binary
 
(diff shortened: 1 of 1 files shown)


freescout-help:
Simply run the following console command before running `update.sh`:
```
alias php='/usr/local/php83/bin/php'
```

Or try this https://www.ionos.com/digitalguide/websites/web-development/update-php-version-and-cancel-php-extended-support/

PR #4308, opened 2024-10-26
Update profile.blade.php - fix typo

DavidAnderson684:
Keep in mind that pull requests should be sent to the `master` branch! See https://github.com/freescout-helpdesk/freescout/wiki/Development-Guide#github-workflow

Fixes a typo: off/of.

Is this setting meant to be used for detecting forwarded messages from a  support agent? (If so, it doesn't work for me).


freescout-help:
Thanks. It has been fixed in the `master` branch.

PR #4316, opened 2024-11-01
Limit customer access

ar45:
Limit customer access to admin or users that have access to mailbox with conversations of customer.

### DIFF
--- a/app/Http/Controllers/ConversationsController.php
+++ b/app/Http/Controllers/ConversationsController.php
@@ -2991,6 +2991,8 @@ class ConversationsController extends Controller
      */
     public function searchCustomers($request, $user)
     {
+        $limitVisibility = config('app.limit_user_customer_visibility') && !$user->isAdmin();
+
         // Get IDs of mailboxes to which user has access
         $mailbox_ids = $user->mailboxesIdsCanView();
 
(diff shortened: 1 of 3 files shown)


freescout-help:
Thanks. This solution does not take one thing into account:

`IN` statement in DB usually has some limit on the number of items you can list in brackets, so if the `select` below returns more customers than `IN` in your DB allows you will get an error.

ar45:
Hm. I tried using WHERE EXISTS but I got a laravel error perhaps due to using laravel 5.5

is the following query ok?

```php
$customers_query->where(function ($query) use ($user) {
                $query->whereRaw('EXISTS (
                    SELECT 1
                    FROM conversations JOIN mailbox_user ON conversations.mailbox_id = mailbox_user.mailbox_id
                    WHERE mailbox_user.user_id = ?
                    AND conversations.customer_id = customers.id
                    LIMIT 1
                )', [$user->id]);
            });
```

freescout-help:
It has been adjusted and added as a single commit: https://github.com/freescout-help-desk/freescout/commit/9a6429ea9761b4e758a9f654165122eeef453613

Thanks for the contribution.

freescout-help:
Instructions:
https://github.com/freescout-help-desk/freescout/wiki/FAQ#is-it-possible-to-have-separate-contactscustomers-per-mailbox

freescout-help:
It should be fixed now.

ar45:
Works as expected. Thank you very much!

Issue #4333, opened 2024-11-11
@fwd not recognised when forwarding from Outlook/Exchange


patrickklaeren:
PHP version: 8.2.20
FreeScout version: 1.8.156
Database: MySQL 10.6.7-MariaDB
Are you using CloudFlare: No
Are you using non-official modules: No, only official

When using `@fwd` as the only text in the body of a forwarded email, it is not designating the conversation to the original email.

I have read the FAQs. I have read related issues #2940, #4036 and #4095.

The email was sent from [redacted]@live.co.uk, and forwarded by patrick.klaeren@[redacted]. The conversation should be created under [redacted]@live.co.uk. Instead it is created under patrick.klaeren@[redacted].

The email headers are as such:

```
Subject: Fw: Help again
Thread-Topic: Help again
Thread-Index: XXX
Date: Mon, 11 Nov 2024 00:02:34 +0000
Message-ID:
	<CWLPXXX@XXX.XXX.PROD.OUTLOOK.COM>
References:
	<PR3PXXX@XXX.XXX.PROD.OUTLOOK.COM>
In-Reply-To:
	<PR3PXXX@XXX.XXX.PROD.OUTLOOK.COM>
Accept-Language: en-GB, en-US
Content-Language: en-US
```

The email body is as such:

```html
<div class="elementToProof" style="font-family: Aptos, Aptos_EmbeddedFont, Aptos_MSFontService, Calibri, Helvetica, sans-serif; font-size: 11pt; color: rgb(0, 0, 0);">
@fwd</div>
<div style="font-family: Aptos, Aptos_EmbeddedFont, Aptos_MSFontService, Calibri, Helvetica, sans-serif; font-size: 11pt; color: rgb(0, 0, 0);">
<br>
</div>
<div id="appendonsend" style="color: inherit;"></div>
<hr style="display: inline-block; width: 98%;">
<div id="divRplyFwdMsg" dir="ltr" style="color: inherit;"><span style="font-family: Calibri, sans-serif; font-size: 11pt; color: rgb(0, 0, 0);"><b>From:</b> Patrick Klaeren <[redacted]@live.co.uk><br>
<b>Sent:</b> Monday, November 11, 2024 00:02<br>
<b>To:</b> Patrick Klaeren <patrick.klaeren@[redacted]><br>
<b>Subject:</b> Help again</span>
<div> </div>
</div>
[...]

PR #4335, opened 2024-11-11
Fixes @fwd forwarding when styles are in HTML #4333

patrickklaeren:
Originally, emails are only stripped of HTML tags using `strip_tags`, whilst this is right, the built in function does NOT account for `<style></style>` bodies.

Given a HTML body that starts with:

```html
<style type="text/css" style="display:none;"> P {margin-top:0;margin-bottom:0;} </style><p>@fwd</p>
```

The previous code:

```php
strtolower(trim(strip_tags($body)))
```

would output: 

```html
 P {margin-top:0;margin-bottom:0;} @fwd
```

Leaving the downstream code unable to see `@fwd`.

This pull request further strips HTML by using `preg_replace`, replace `<style>*</style>` from the aforementioned HTML, outputting the correct:

```
@fwd
```

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -729,7 +729,9 @@ class FetchEmails extends Command
                 && !$user_id && !$is_reply && !$prev_thread
                 // Only if the email has been sent to one mailbox.
                 && count($to) == 1 && count($cc) == 0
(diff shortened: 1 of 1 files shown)


freescout-help:
Thanks. It has been fixed using `\Helper::stripTags()` function: https://github.com/freescout-help-desk/freescout/commit/8d65bdf0bf6f8a042f4df98c8e0b8064c7d6e8d6

PR #4340, opened 2024-11-12
Added filter so user can override default conversation sorting

mahavirn:
## Description
Added filter so user can override default conversation sorting. Currently there was no way to change the default sorting.

If anyone wants to override the sorting, they can simply use:

```
Eventy::addFilter('conversations.default.sorting', function() {
  $result = [
     'sort_by' => 'date|subject|number',
     'order' => 'desc|asc'
  ];
  return $result;
}, 20, 1);
```

You can use sort_by as `date` OR `subject` OR `number` And the Sorting order can be `desc` or `asc`;

### DIFF
--- a/app/Conversation.php
+++ b/app/Conversation.php
@@ -2221,6 +2221,8 @@ class Conversation extends Model
             'order' => 'desc',
         ];
 
+        $result = \Eventy::filter('conversations.default.sorting', $result);
+
         if (
             !empty($request->sorting['sort_by']) && !empty($request->sorting['order']) &&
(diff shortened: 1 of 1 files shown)


mahavirn:
CC: @freescout-help @freescout-help-desk

freescout-help:
It has been added as a new commit: https://github.com/freescout-help-desk/freescout/commit/21053200f00286d78394b48df2eca81ddf1acc2d

Notice the updated filter name: `conversations.table_sorting`

PR #4346, opened 2024-11-13
Croatian translation

shunkica:
Update Croatian translation.

Please check if this identifier is correct?

`This reply will go to the customer. :%switch_start%Switch to a note:switch_end if you are replying to :user_name.`

Shouldn't it be 

`This reply will go to the customer. :%switch_start%Switch to a note:%switch_end% if you are replying to :user_name.`


freescout-help:
It will be fixed.

PR #4396, opened 2024-12-04
Action toolbar changes and added a new filter

Lars-:
This change makes it easier to update the action toolbar and dropdown. This way, new modules can rearrange any icon in the toolbar and dropdown.

Also added a filter for the move_conv modal.


freescout-help:
Some adjustments are needed:

- The logic of showing "Delete" and "Delete Forever" buttons got a bit broken. These buttons should be shown in the toolbar on large screens and moved to the dropdown on small screens.
- Rename `ConversationActions.php` into `ConversationActionButtons.php`
- Rename `conversation.actions` hook into `conversation.action_buttons`

Lars-:
All changes have been processed. Maybe `conversation.action_buttons` is a bit tricky since it's an action (already exists before my changes) and a filter in the new helper class. But it should work.

freescout-help:
Ok. So better to use `conversation.get_action_buttons` for the hook in the helper class.

Also Eventy can be addressed as `\Eventy`, so this is not needed: `use Eventy;`

Lars-:
Changed!

Issue #4400, opened 2024-12-04
Look at ditching the `ps` requirement


jdevalk:
PHP version: any
FreeScout version: latest
Database: any
Are you using CloudFlare: Yes

I'm running into the fact that on a SiteGround GoGeek plan, which in itself is perfectly fine to run FreeScout on, `ps` is not available. In looking into where `ps` is used, I find it [here](https://github.com/freescout-help-desk/freescout/blob/dist/app/Misc/Helper.php#L1453), and [here](https://github.com/freescout-help-desk/freescout/blob/dist/app/Http/Controllers/SystemController.php#L100). 

I'm wondering if it wouldn't be possible to keep track of running processes inside the app better, so `ps` wouldn't be needed?

jdevalk:
Oh and to be clear: I'm willing to fund the development of this to some extent if needed.

jdevalk:
I've found a workaround which makes it "sort of" work: I'm firing off a regular cron job that does

```php artisan freescout:send-monitor```

which seems to fix the fact that emails aren't sending.

**Comment by remcotolsma** [comment]
I'm running in the same issue, system status also reports:

<img width="1118" alt="Scherm­afbeelding 2024-12-10 om 09 40 09" src="https://github.com/user-attachments/assets/1b2dea4b-7a13-4204-ac91-b48292e4385b">

> `queue:work`
> Laatste run: ? Laatste succesvolle run: ? Probeer de cache te legen om de start van het commando te forceren.

I found the following issue about this:

- https://github.com/freescout-help-desk/freescout/issues/4329

Checked https://github.com/freescout-help-desk/freescout/wiki/Background-Jobs and executed the following command: 

```
php artisan freescout:check-requirements
```

```
Error, do this: mount -t proc proc /proc
Error, do this: mount -t proc proc /proc
Error, do this: mount -t proc proc /proc
Error, do this: mount -t proc proc /proc
PHP Version
 8.2.26........................ OK
PHP Extensions
 OpenSSL....................... OK
 PDO........................... OK
 Mbstring...................... OK
 Tokenizer..................... OK
 JSON.......................... OK
 XML........................... OK
 IMAP.......................... OK
 GD............................ OK
 fileinfo...................... OK
 ZIP........................... OK
 iconv......................... OK
 cURL.......................... OK
 DOM........................... OK
 libxml........................ OK
 intl.......................... OK
 pcntl (console PHP)........... OK
Error, do this: mount -t proc proc /proc
[...]

PR #4414, opened 2024-12-10
feat: replace ps requirement with native php solution

Lars-:
Possible fix for #4400


freescout-help:
Thanks for the effort.

https://github.com/freescout-help-desk/freescout/issues/4400#issuecomment-2532361018

PR #4551, opened 2025-02-06
Add tests for the `\App\Misc\Mail::replaceMailVars()` method.

thomasplevy:
Hello team,

My team has decided that we require merge code fallbacks so I'm queuing up a rewrite of the method that handles replacing mail vars.

I am opening this PR as a draft in preparation for the rewrite, my thinking is that I want to guarantee that this set of tests works before and after my rewrite so that the implementation can be verified to be backwards compatible and free of regression.

This PR adds unit tests with (largely) mocked data objects in order to best control the expected input and output of the `replaceMailVars()` method. 

I believe this set of tests covers all possible input values and merge codes present in the core codebase.

Thanks!

### DIFF
--- a/tests/Unit/MailVarsTest.php
+++ b/tests/Unit/MailVarsTest.php
@@ -1,0 +1,290 @@
+<?php
+
+namespace Tests\Unit;
+
+use App\Misc\Mail;
+use Generator;
+use Tests\TestCase;
+
+class MailVarsTest extends TestCase
+{
+    /**
+     * Retrieves an array of mocked data that can be passed to the {@see \App\Misc\Mail::replaceMailVars()} method's $data parameter.
+     *
+     * @return array
+     */
+    protected static function fakeMailData(): array
+    {
+        return [
+            'conversation' => (object) [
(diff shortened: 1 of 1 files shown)


freescout-help:
Can you please:
- Rename it into `MailVarsTest`
- Make sure the code is compatible with PHP 7.1

thomasplevy:
> Rename it into MailVarsTest

Updated.

> Make sure the code is compatible with PHP 7.1

I'm having a little trouble with this one, I'm working through the issues step by step but I'd like to know how you'd like me to proceed:

1. I've downgraded to composer 2.2 (the last version of composer that supports PHP 7.1)
2. Installing composer deps with `composer install --ignore-platform-deps`
3. Can't run PHP Unit 9.5x on PHP 7.1 so I've updated the requirement in composer.json to v7.5.20 (the last PHPUnit version that supports PHP 7.1

After that I did have a problem with PHP 7.1 in the tests in this PR, I fixed those, sorry about that!

I have opted not to commit the composer changes as they'd create issues with later PHP versions.

Anything else you'd like me to do here please let me know

Issue #4473, opened 2025-01-13
Placeholder Variables Displayed in Email Body When Undefined


KuroshS:
When using saved replies in FreeScout, if a placeholder variable (e.g., `{%customer.firstName%}`) has no assigned value, the variable itself is displayed in the email body rather than being omitted or replaced with a default value.

**Example:**  
Template:  
`Hello {%customer.firstName%}`  

Output (if `firstName` is undefined):  
`Hello {%customer.firstName%}`  

Is it possible to replace the undefined variable with a fallback value, like in Helpscout, or omit entirely?

freescout-help:
Unfortunately there is no such plans for now.

Issue #4139, opened 2024-07-21
Fallback for variables


PoeHaH:
Hi,

I love the "variables" feature in saved replies. I'd like to implement "fallback" behavior just like HelpScout. Their variable looks like this: 

Hi `{%customer.firstName,fallback=there%}`,

If `customer.firstName` is not available, "there" will be inserted instead.

I'd love to implement similar behavior, but there's no filter to alter the output from variables? Would you be willing to include it?

Thanks!

freescout-help:
It would be better to implement this in the core of the application.

PoeHaH:
Can I do something to make this happen?

freescout-help:
Yes, you can implement it and submit a pull request.

MarianBMI:
hi @freescout-help I totally understand @PoeHaH . This feature should be set by default. Why not setting it?

PR #4552, opened 2025-02-06
Refactor `\App\Misc\Mail::replaceMailVars()` and add merge var fallback capabilities

thomasplevy:
See related #4551 

The primary purpose of this PR is to introduce a method for providing fallback strings for any merge code used in the application.

I see a few issues for this (closed) and noted in one of them that help may be accepted: #4139, #4473

My team has determined that we really do need fallback options (we're coming to Freescout after years on Helpscout)

Since the application is largely modeled after Helpscout's feature set, this PR implements fallbacks largely as Helpscout does: `{%customer.fullName,fallback=there%}`

This implementation allows providing a fallback for every available merge code without discrimination.

The strings provided following the equals sign will be used if the mapped data object's value is empty.

If a fallback is provided with no value, eg `{%customer.fullName,fallback=%}` an empty string will be used resulting in the merge code itself being removed from the final text regardless of the value provided for `$remove_non_replaced`. __I need a little bit of feedback on this particular decision as it's unclear to me by reading through the code alone what the purpose of this argument is. I may have made a bad assumption here and created an issue. If that's the case please let me know and I'll update the code.__

This PR also introduces new filters which will, hopefully, make it "easier" for developers to add custom merge vars to the application.

### Note on premium modules 

I have access to a few custom modules that utilize the pre-existing `mail_vars.replace` filter, the CRM and Custom Fields modules.

Based on the way the pre-existing filter works these merge codes should continue to function as expected however if a user adds a fallback value they will stop working.

In order to add support for fallbacks on these custom fields the modules would need to be updated to use the new filters introduced into this PR. For the Custom Fields module that would look something like this (untested):

```php
// Add custom fields to available mail vars.
\Eventy::addFilter('mail_vars.vars`, function(array $vars, array $data): array {
   if (! empty($data['conversation]) {
       foreach(CustomField::getMailboxCustomFields($data['conversation']->mailbox_id) as $field) {
           $vars[] = $field->getNameEnconded();
       }
       return $vars
       }
   }
   return $vars;
}, 20, 2);

// Replace custom field merge var values.
foreach(CustomField::getMailboxCustomFields($data['conversation']->mailbox_id) as $field)
     $name = $field->name;
     \Eventy::addFilter('mail_vars.replacement.' . $field->getNameEncoded(), function($val, $data, $fallback) use($name)
     if (! empty($data['conversation']) {
         return CustomField::getByName($name, $data['conversation']->mailbox_id, $data['conversation']->id); // This is a made up function that would retrieve a single field value by name instead of an array of all field values.
     }, 20, 3);
}

### CODE REVIEW

**Comment by freescout-help** [comment]
1) The solution should not break existing vars when `,fallback=""` is added to them, the `fallback` should be simply ignored in such cases.

2) These things are not secure as may allow to get access to any property of an object:
```
$data[$dataPath[0]]->{$dataPath[1]}
```

3) The algorithm itself seems seems overcomplicated. How it can be done:

1. Return the code to its original state.
2. After the `$vars` array is prepared in `replaceMailVars()` use `preg_match_all` to get all the {%...%} placeholders from the text.
3. Go through the `$vars` array and modify its keys and values according to the data obtained by `preg_match_all`.

thomasplevy:
@freescout-help thanks for the feedback, I appreciate it.

I'm going to assume that by providing feedback that does mean you're interested in this feature so that excites me. Thanks

I'll review everything in more detail and update the PR next week. I don't think I'll be able to get a reworking with your feedback in place before the weekend.

Thanks again,

freescout-help:
Yes, it would be good to have this feature.

thomasplevy:
I am not abandoning these changes but have not had time to work on them last week or so far this. Just checking in to let you know I'm not disappearing just busy with other priorities. Thanks for your patience!

freescout-help:
Ok. No hurry. Just create a new PR when it's ready.

PR #4607, opened 2025-02-26
Use keyMap Enter->insertParagraph in editor as it insert divs

anxodio:
Fixes https://github.com/freescout-help-desk/freescout/issues/3811

In earlier verisons of freescout, not all 'p' where replaced with 'div' so the insertParagraph method didn't work. But now it's well replaced (I checked it and marked every change with a comment), so it's safe to use insertParagraph again.

So this PR just adds the comments, remove the forced shift in enter, and puts again the keymap.

With this PR now it's easy to go out from code/citation blocks (just press enter, if you wanna a new line inside you have to press shift+enter), keeping divs as a separator.

It also keep other expected behaviours like pressing shift+enter in a list to add a subline.

### DIFF
--- a/public/js/summernote/summernote.js
+++ b/public/js/summernote/summernote.js
@@ -1610,6 +1610,7 @@ var dom = {
     /** @property {String} blank */
     blank: blankHTML,
     /** @property {String} emptyPara */
+    // Freescout uses div instead of p to searate paragraphs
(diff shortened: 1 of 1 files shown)


freescout-help:
https://github.com/freescout-help-desk/freescout/issues/4909

PR #4645, opened 2025-03-13
fix: remove absolute positioning on customer-social-profile links element in customer profile

### LINKED ISSUE #4644: Visual issues on customer profile when the customer has more than 12 social profiles and/or links
If a customer has more than 12 links they start to overlap with the customer's name/email/etc... in the customer's profile sidebar

Image

In this instance the customer has 30 links.

It's a bit excessive but the data exists for some of our customers so we'd like to retain it instead of deleting the information.

I've fiddled with some quick fixes for this and overwriting the CSS to remove the `absolute` positioning on the `.customer-social-profiles` div "fixes" the problem by moving the links inline under the email address. I also added a little padding.

Image

This is acceptable to me but I don't know if it's a permanent solution and wanted to report the issue in case you might want to resolve it upstream.

I'd be happy to open a PR to address the issue if you're interested in my proposed solution

Thanks

thomasplevy:
resolves visual overlap issues when a customer has 12+ links on their profile

Fixes #4644

### DIFF
--- a/public/css/style-rtl.css
+++ b/public/css/style-rtl.css
@@ -31,11 +31,6 @@ body.locale-fa {
     position: relative;
 }
 
-.customer-social-profiles {
-    right: 92px;
-    left: auto;
-}
-
 .accordion .panel-title > a .caret {
     float: left;
 }
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -1233,9 +1233,7 @@ https://github.com/freescout-help-desk/freescout/issues/4354
     margin: 0 0 4px 0;
 }
 .customer-social-profiles {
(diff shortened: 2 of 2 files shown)


freescout-help:
Thanks for the implementation. We've checked it. Unfortunately it does look very good in reality. So we'll just wait for some a bit more elegant solution to appear.

thomasplevy:
sure this was just a quick solution, in the issue I didn't mean to indicate i'd do anything other than address it like my screenshot. 

if you have some direction on what a "more elegant" solution might look like I would be happy to work on that but without more direction I'm not going to guess at a better option for today.

Thanks

PR #4691, opened 2025-04-07
fix: it is "Cloudflare", not "CloudFlare"

BenjaminEHowe:
[Cloudflare dropped their capital "F"](https://blog.cloudflare.com/time-for-an-update/) so I thought I'd quickly update the FreeScout codebase to match. Thanks to the maintainers for your hard work!

### DIFF
--- a/.github/ISSUE_TEMPLATE/general_help_request.md
+++ b/.github/ISSUE_TEMPLATE/general_help_request.md
@@ -21,5 +21,5 @@ Still here? Well clean this out and go ahead :)
 PHP version: 
 FreeScout version: 
 Database: MySQL / PostgreSQL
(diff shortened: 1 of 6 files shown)


freescout-help:
Thanks for the pull request. We'll just leave it as it is for now.

BenjaminEHowe:
Hmm – I'm intrigued as to why we wouldn't fix this bug. It's obviously not a big deal, but a bug is a big...

Would you prefer it if I submitted an alternative PR which only fixes the user-facing instances of this and leaves things as-is internally / in the comments / etc. ?

freescout-help:
It's not so important. Tomorrow they may choose some other spelling.

PR #4693, opened 2025-04-08
add mail merge var fallback capabilities

thomasplevy:
This is a follow-up to my initial PR: https://github.com/freescout-help-desk/freescout/pull/4552

This PR simplifies the methodology from my initial PR based on your feedback.

---

This PR introduces a method for providing fallback strings for any merge code used in the application.

I see a few issues for this (closed) and noted in one of them that help may be accepted: #4139, #4473

My team has determined that we really do need fallback options (we're coming to Freescout after years on Helpscout)

Since the application is largely modeled after Helpscout's feature set, this PR implements fallbacks largely as Helpscout does: `{%customer.fullName,fallback=there%}`

This implementation allows providing a fallback for every available merge code without discrimination.

The strings provided following the equals sign will be used if the mapped data object's value is empty.

If a fallback is provided with no value, eg `{%customer.fullName,fallback=%}` an empty string will be used resulting in the merge code itself being removed from the final text regardless of the value provided for `$remove_non_replaced`. __I need a little bit of feedback on this particular decision as it's unclear to me by reading through the code alone what the purpose of this argument is. I may have made a bad assumption here and created an issue. If that's the case please let me know and I'll update the code.__

### A new filter

This PR also introduces a new filter, `mail_vars.replace_after_fallback` which is identical to `mail_vars.replace` except that it runs **after** the `$vars` array is updated with fallback data.

I'm not certain this is a necessary filter but I thought it would be nice to include a way for devs to interacting with the array after the interal fallback logic runs.

### Note on premium modules 

I have access to a few custom modules that utilize the pre-existing `mail_vars.replace` filter, the CRM and Custom Fields modules.

Based on the way the pre-existing filter works these merge codes will continue to function as expected without modification to the module code. Fallbacks will start to work immediately as well.

I need some help testing and confirming that this is true though across all modules though. I'm assuming implementation is similar across other custom modules that implement merge vars but that's just an assumption.


Stiofan:
Well done, much needed, this was one of the only things I was not happy about when moving over :)

PR #4697, opened 2025-04-11
Adds functionality to mark notifications as unread

Lars-:
Allows users to mark notifications as unread.
Improves user experience by providing a way to revisit notifications.

Refreshes notification counter periodically without user interaction.


freescout-help:
Thanks. But it has been decided not to merge it. Usefulness of refreshing Notifications counter by sending ajax request every 2 minutes is questionable. And clicking the "Mark as unread" link inside notification block which is clickable itself may and will cause troubles especially on mobile devices.

PR #4715, opened 2025-04-23
Added actions and filters

Lars-:
I've changed my modifications from last week to actions and filters so the desired functionality can be built in a custom module.


freescout-help:
Notice that `users.ajax.response` hook has been renamed into `users.ajax.response_default`.

PR #4733, opened 2025-05-04
Take in account possible spaces when handling forwarded emails

Pierstoval:
This will take in account other cases, I just encountered a forward subject with a syntax like `TR : *initial subject*` (note the presence of a space character)


freescout-help:
Thanks. Added as https://github.com/freescout-help-desk/freescout/commit/0b6a314f013d9a1f3af06529e78b207200ddc5e0

PR #4797, opened 2025-05-21
Revise setCookie per best practices

presswizards:
This edit fixed my Dark Mode cookie issue in Edge browser:

- Lower-cased all cookie attributes, as browser expects them lowercase.
- Set default path always.
- Added the Secure flag automatically if using samesite=None.
- Explicitly add expiration as UTC string.
- Fixed the calculation for expiration.


freescout-help:
Thanks. Already applied.

PR #4811, opened 2025-05-28
Add filter to allow modules to override outgoing email "From" name in…

pv8888888:
… ReplyToCustomer.php

This change introduces a call to the Eventy filter 'email.reply_to_customer.from_name' before setting the "From" name in outgoing customer replies.

- Allows modules to dynamically override the "From" name for outgoing emails.
- If no module uses the filter, the default behavior is unchanged.

Backward compatible.


freescout-help:
1) Make sure to submit the PR into `master` branch: https://github.com/freescout-help-desk/freescout/wiki/Development-Guide#github-workflow

2) Try not to add/remove spaces/tabs to let us see what actually changed in the code.

PR #5151, opened 2026-01-12
Version 2 - #4644 Removed CSS Style on Social Profiles / Added additional class to social profiles blade

raramuridesign:
This suggestion places the social profiles below the emails, it looks cleaner and adopts the spacing using existing classes.

Addresses issues #4644

CSS file correctly formatted and css for social profiles class disabled

Please consider this addition. Thanks M

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -1475,6 +1475,11 @@ https://github.com/freescout-help-desk/freescout/issues/4354
     top: 0;
     left: 92px;*/
 }
+.social-profiles-few{
+    position: absolute;
+    top: 0;
+    left: 92px;
+}
 
 .customer-social-profiles a {
     font-size: 17px;
--- a/resources/views/customers/profile_snippet.blade.php
+++ b/resources/views/customers/profile_snippet.blade.php
@@ -35,12 +35,25 @@
 	        @endforeach
 		</ul>
 		<div class="customer-extra">
-			@if ($customer->getSocialProfiles() || $customer->getWebsites())
-				<div class="customer-section customer-social-profiles">
-					@foreach ($customer->getWebsites() as $website)
+			@php
+				$socialProfiles = $customer->getSocialProfiles();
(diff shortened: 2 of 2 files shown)


freescout-help:
Unfortunately we don't see much difference in the way social profiles are displayed.
Maybe something also needs to be changed in styles?

raramuridesign:
@freescout-help 
When you say something needs to be changed in styles, what do you mean by this? 
The CSS which was changed is this .customer-social-profiles - the style elements were removed. 

We could do the following
1. add a background colour to the social media area or
2. if possible have different icons based on the profile added, but currently this is not handled in anyway

If you can provide your thoughts, then I can look at this.
M

freescout-help:
This is how it looks. Is it supposed to be like this?

raramuridesign:
@freescout-help it should look like this

image

freescout-help:
Thanks. This solution is a bit space-consuming when there is just 1 or 2 social profiles.

raramuridesign:
@freescout-help The following has been added

1- added 2 new classes 
.social-profiles-few  - this has been added to style.css
.social-profiles-many

2. In profile_snippet.blade.php I have added a check for 10 profiles - if less than or equal to 10, then it will use the class .social-profiles-few - if more than 10 then it will use .social-profiles-many 

The placement of -few is the original placement.
If its -many, this goes to the placement as suggested in the screenshot.

I trust this makes sense. Send me your comments
M

freescout-help:
Thanks for ideas. Fixed via the scroll this way: https://github.com/freescout-help-desk/freescout/commit/bcee603c02ac8e2f465ea76773ce228462a008d2

raramuridesign:
@freescout-help If you view the screenshot this is not elegant- I understand you dont want to move it, but considering when there are more profiles this looks really messy. The solution I proposed does sort this out. 

image

Hope you consider this.

PR #5198, opened 2026-02-06
Fix To header name mismatch after requester change (#3149)

### LINKED ISSUE #3149: Customer receives email with original customer name despite customer change
Hello all,

if I change the email address of the customer in an open ticket, the first and last name (display name) of the previous customer remains, even if the email changes.

So if I have a ticket open from the customer "Ben Wight" with the email address "[ben.wight@mail.com](mailto:ben.wight@mail.com)" in which "Mona Wight" is in CC with "[mona.red@mail.com](mailto:mona.red@mail.com)" and I want to send a reply in the ticket to "Mona Red", I click on the reply button and select "[mona.red@mail.com](mailto:mona.red@mail.com)". In FreeScout this all looks neat and correct.

In the mail client of "Mona Red", however, an e-mail comes in with the following recipient: "Ben Wight [mona.red@mail.com](mailto:mona.red@mail.com)".

Is this a deliberate behaviour or a bug?

Many thanks for your help!

Best regards
Lukas

patthewebrat:
## Problem

When a conversation's requester is changed from Customer A to Customer B, outgoing reply emails use the wrong name in the `To` header:

```
To: Customer B <a@example.com>   ← wrong: Customer B's name, Customer A's email
To: Customer A <a@example.com>   ← correct
```

This happens because `SendReplyToCustomer.php` line 280 always uses `$this->customer->getFullName()` — the conversation's current requester — regardless of which customer owns `$this->customer_email`.

## Fix

Look up the customer record for the actual email being sent to via `Customer::getByEmail()` and use that customer's name. This method already exists (`app/Customer.php:1246`) and is already used in the same file at line 249.

When no requester change has occurred, `Customer::getByEmail()` returns the same customer, so behaviour is unchanged for normal replies.

## Testing

1. Change a conversation's requester from Customer A to Customer B
2. Reply to the conversation
3. Inspect the outgoing email `To` header — should show `Customer A <a@example.com>`
4. Verify normal reply flow (no requester change) is unaffected

Fixes #3149

### DIFF
--- a/app/Jobs/SendReplyToCustomer.php
+++ b/app/Jobs/SendReplyToCustomer.php
@@ -282,7 +282,16 @@ class SendReplyToCustomer implements ShouldQueue
         if (count($to_array) > 1) {
             $to = $to_array;
         } else {
-            $to = [['name' => $this->customer->getFullName(), 'email' => $this->customer_email]];
+            // Look up the correct customer name for the email being sent to.
+            // $this->customer may not match $this->customer_email after a requester change.
+            $to_customer_name = $this->customer->getFullName();
+            if ($this->customer_email) {
+                $email_customer = Customer::getByEmail($this->customer_email);
+                if ($email_customer) {
+                    $to_customer_name = $email_customer->getFullName();
+                }
+            }
+            $to = [['name' => $to_customer_name, 'email' => $this->customer_email]];
         }
 
         // If sending fails, all recipiens fail.


freescout-help:
Thanks.

1) Pull requests should be sent to the `master` branch.
2) Check `$this->customer->fresh()`. Maybe it will solve the problem.

patthewebrat:
Hey there,

`$this->customer->fresh()` suggests a staleness related issue, however I have tried this fix and it doesn't work.

This is because the fundamental problem is that $this->customer->getFullName() gets the current requester, not the recipient of the email.

As such, when the To drop down field appears, if you select a different recipient to the requester, the name in the email headers will always be the requester's name (even though the email belongs to the selected recipient).

Sorry about the branch error, I will create PR against master for your review.

Thanks,

Pat

PR #5211, opened 2026-02-12
Do not throw exceptions on E_DEPRECATED warnings to fix compatibilty with newer PHP versions

RisingOpsNinja:
Currently E_DEPRECATED warnings are handled like fatal errors in Freescout and exceptions are thrown. This leads to the fact that Freescout no longer works properly after major PHP updates, although they are actually only warnings.

This change silences E_DEPRECATED completely and solved issues with PHP 8.5 for my deployment. But other solutions may also be possible e.g.: don't set error_reporting at all and depend on the php setting or filter out E_DEPRECATED but still log them.

There were a lot issues and changes in the past due to this behavior: e.g.

https://github.com/freescout-help-desk/freescout/issues/5177
https://github.com/freescout-help-desk/freescout/issues/5134
https://github.com/freescout-help-desk/freescout/issues/5114
https://github.com/freescout-help-desk/freescout/issues/4979
https://github.com/freescout-help-desk/freescout/issues/4941
https://github.com/freescout-help-desk/freescout/issues/4857
https://github.com/freescout-help-desk/freescout/issues/4848

### DIFF
--- a/overrides/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
+++ b/overrides/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
@@ -29,7 +29,7 @@ class HandleExceptions
     {
         $this->app = $app;
 
-        error_reporting(-1);
+        error_reporting(E_ALL & ~E_DEPRECATED);
 
         set_error_handler([$this, 'handleError']);
 


freescout-help:
The thing is that exceptions are needed to make FreeScout compatible with each new PHP version. It would be better to share error which you are facing and they will be fixed with time.

RisingOpsNinja:
I created issues for the errors I faced:
https://github.com/freescout-help-desk/freescout/issues/5215
https://github.com/freescout-help-desk/freescout/issues/5214
https://github.com/freescout-help-desk/freescout/issues/5213

RisingOpsNinja:
> The thing is that exceptions are needed to make FreeScout compatible with each new PHP version. It would be better to share error which you are facing and they will be fixed with time.

The exception just prevent that Freescout works with new PHP versions that would work perfectly fine. It is good that deprecation are addressed early of course, but it is a bad user experience when the application breaks because of it. Deprecations are meant to be warnings.

 e.g.  _implicitly nullable parameter_ types is planned to be removed only in PHP 9 which has no release date yet. But it broke Freescout multiple times already.

>  And remove support for implicitly nullable types in PHP 9. 

https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

PR #5232, opened 2026-02-21
fix: authorization bypass in conversation_change_customer action

theeggorchicken:
### Vulnerable Lines

**File:** [`app/Http/Controllers/ConversationsController.php#L1848-L1851`](https://github.com/freescout-helpdesk/freescout/blob/dist/app/Http/Controllers/ConversationsController.php#L1848-L1851)

```php
// Lines 1848-1851 (before fix)
$conversation->changeCustomer($customer_email, null, $user);   // runs even when permission check fails

$response['status'] = 'success';
\Session::flash('flash_success_floating', __('Customer changed'));
```

The permission checks above these lines (lines 1838-1845) correctly detect unauthorized access and set `$response['msg'] = 'Not enough permissions'`, but the `changeCustomer()` call sits outside the guard and executes unconditionally.

### What could happen

Any authenticated FreeScout user - even one with no mailbox access - can reassign a conversation's customer by POSTing to `/conversation/ajax` with `action=conversation_change_customer`. The server returns `{"status":"success","msg":"Not enough permissions"}`, acknowledging the permission failure in the message while still performing the database mutation. This lets a low-privilege user (role=1, no mailbox access) silently transfer conversations to arbitrary customers.

### How to verify

1. Create a FreeScout user with `role=1` (ROLE_USER) and **no** mailbox access
2. Log in as that user and send:

```bash
curl --trace-ascii /tmp/trace.txt -X POST \
  -H "X-CSRF-TOKEN: <token>" \
  -H "X-Requested-With: XMLHttpRequest" \
  -H "Cookie: laravel_session=<session>" \
  -d "action=conversation_change_customer&conversation_id=1&customer_email=attacker@evil.com&_token=<token>" \
  http://localhost:9090/conversation/ajax
```

3. Response returns `{"status":"success","msg":"Not enough permissions"}`
4. Check the database - the conversation's `customer_email` has changed to `attacker@evil.com` despite the "Not enough permissions" message

Full HTTP trace with `curl --trace-ascii`: [evidence gist](https://gist.github.com/theeggorchicken/09b4abd94b5994dd74ed7584dff8dd97)

### Visual proof

**Before exploit** - conversation belongs to "Original Customer <original@customer.com>":

Before

**After exploit** - same conversation now shows "Attacker Evil <attacker@evil.com>", and the activity log records the unauthorized change:

After

### What this PR does

Wraps the `changeCustomer()` call, the success status assignment, and the `Session::flash` inside the existing `if (!$response['msg'])` guard pattern that other actions in this switch block already use. The mutation only runs when all permission checks pass (i.e., `$response['msg']` is still empty).

```php
if (!$response['msg']) {
    $conversation->changeCustomer($customer_email, null, $user);

    $response['status'] = 'success';
    \Session::flash('flash_success_floating', __('Customer changed'));
}
```

### Tests included

`tests/Feature/ConversationChangeCustomerTest.php` - two cases:

- **testUnauthorizedUserCannotChangeCustomer**: Creates a user with `role=1` and no mailbox access, fires the change-customer request, asserts the response contains "Not enough permissions", and confirms the database row is unchanged. This test would fail before this fix.
- **testAuthorizedUserCanChangeCustomer**: Creates an admin with mailbox access, fires the same request, asserts success with empty message, and confirms the customer was updated. Regression check that normal functionality still works.

### Evidence

- [curl --trace-ascii dump](https://gist.github.com/theeggorchicken/09b4abd94b5994dd74ed7584dff8dd97) - full HTTP request/response showing the exploit
- Before/after screenshots committed in `evidence/screenshots/` (see images above)


freescout-help:
Thanks. Added as https://github.com/freescout-help-desk/freescout/commit/417e25dd44ca1869dd103697aab8bbdeb2db43f8

1) Pull requests should be sent to the `master` branch: https://github.com/freescout-help-desk/freescout/wiki/Development-Guide#github-workflow
2) The `tests/Feature/ConversationChangeCustomerTest.php` was not really working and has been fixed.

PR #5252, opened 2026-03-04
Reduced Cc and Bcc in languages where the text does not fit nicely

il2020:
https://github.com/freescout-help-desk/freescout/issues/5247


freescout-help:
Thanks. But PR contains a lot of unneeded commits.

You can share your update as a patch: https://stackoverflow.com/questions/5159185/how-to-create-a-git-patch-from-the-uncommitted-changes-in-the-current-working-di

PR #5274, opened 2026-03-14
Fix accessibility issues with magic checkboxes used in the conversations table

nktech1135:
This fixes the issue where checkboxes were not recognized by screen readers in the conversations table. It may also fix other areas  #@where magic checkboxes are used.


freescout-help:

Applied this way: https://github.com/freescout-help-desk/freescout/commit/0565ba97b74ecef26fcc742d1dcfbeb4d6698f5e

PR #5327, opened 2026-04-10
Update Thread.php

ithomeheavenly:
Add a hook in case you want to bypass HTML sanitization in a particular module. I’ve encountered situations where emails didn’t display properly due to this sanitization—specifically, the HTML buttons in the notifications sent to me by Manomano were disappearing. That’s why I need to add this hook so that future updates don’t break my current code.

### DIFF
--- a/app/Thread.php
+++ b/app/Thread.php
@@ -307,6 +307,11 @@ class Thread extends Model
         if ($body === null) {
             $body = '';
         }
+        //Skip cleaning the body and remove it as is
+		$skipCleanBody = \Eventy::filter('thread.skip_clean_body', false, $this, $body);
+		if ($skipCleanBody) {
+			return $body;
+		}
 
         // Change "background:" to "background-color:".
         // https://github.com/freescout-helpdesk/freescout/issues/2560


ithomeheavenly:
errorsolution
I'm attaching a photo so you can see what was going on

freescout-help:
It would make the system less secure. Buttons disappear for you as they are no allowed in Purifier config: https://github.com/freescout-help-desk/freescout/blob/dist/config/purifier.php

We'd recommend to find some workaround or replace `<button>` with something else.

ithomeheavenly:
I understand what you're saying. Ultimately, this mailbox connects directly to Manomano, so there wouldn't be any security issues from there. Plus, in my custom Manomano module, I've already included a cleanup feature for Manomano and restricted access to only the Manomano mailbox.

I understand that it’s somewhat risky and you don’t want to set it to master. For now, I’ll keep it that way in my Freescout installation; in the future, I’ll look for a safer alternative so I don’t have to worry about re-adding the hook to the core with every update.

Thanks

PR #5353, opened 2026-04-16
Fix RTL: onoffswitch clicks no longer scroll the page to the left

### LINKED ISSUE #5352: [Bug] RTL: clicking on/off switch toggles scrolls the whole page to the left
### Describe the bug

In RTL locales (Hebrew, Arabic, Persian, etc.), clicking any on/off switch (e.g. on Settings → General, Settings → Alerts, Mailbox → Auto Reply, Mailbox → Connection Settings) causes the entire page to scroll/jump to the left. The user has to scroll back to the right to continue working.

### Root cause

In `public/css/style.css` (around line 3213), the hidden checkbox uses off-screen positioning:

```css
.onoffswitch-checkbox {
    position: absolute;
    left: -9999px;
}
```

When the `<label>` is clicked, focus is moved to the hidden `<input>` sitting at `left: -9999px`. Browsers then scroll the viewport horizontally to bring the focused element into view. In LTR this often goes unnoticed, but in RTL the scroll origin is flipped, so the page visibly jumps to the far left on every click.

A secondary issue: the knob uses `right: 19px` → `right: 0` for the unchecked/checked states, which is not mirrored for RTL, so the knob slides in the wrong direction on RTL pages.

### Steps to reproduce

1. Set locale to Hebrew/Arabic/Persian (any RTL language).
2. Open Settings → General (or any page with an on/off switch).
3. Click any toggle.
4. Observe the whole page jumping to the left.

### Expected behavior

Toggling the switch should not affect the page scroll position. The knob should slide in the correct (mirrored) direction in RTL.

### Environment

- **FreeScout version:** 1.8.215
- **Browser:** Reproducible in Chrome, Firefox, and Edge
- **OS:** Any

### Suggested fix

Replace the off-screen hiding with an accessible "visually hidden" pattern, and mirror the knob position for RTL. 

**Patch for `public/css/style-rtl.css`:**

```css
/* Hide the checkbox without off-screen positioning to prevent the browser 
   from horizontally scrolling the page when the label is clicked. */
.onoffswitch-checkbox {
    left: auto;
    right: auto;
    width: 1px;
    height: 1px;
    opacity: 0;
    clip: rect(0 0 0 0);
    overflow: hidden;
    margin: 0;
    padding: 0;
    border: 0;
}

/* Mirror the knob for RTL: in LTR it slides from right:19px to right:0, 
   so in RTL it should slide from left:19px to left:0. */
.onoffswitch-label:before {
    left: 19px;
    right: auto;
}

.onoffswitch-checkbox:checked + .onoffswitch-label:before {
    left: 0;
    right: auto;
}
```

Alternatively, the hiding technique can be applied directly in `public/css/style.css` so LTR benefits from it too (the focus-scroll issue technically exists in LTR as well, it's just less visible).

Happy to open a PR if you'd like.

uhbudrs:
Fixes #5352.

## Summary

In RTL locales, clicking any on/off switch (Settings → General, Alerts, Mailbox → Auto Reply, etc.) caused the whole page to jump to the left because the hidden checkbox sits at `left: -9999px` and browsers scroll horizontally to bring the focused element into view.

## Changes

**public/css/style-rtl.css:**

- **Hide the checkbox without off-screen positioning.** Replace `left: -9999px` positioning (inherited from style.css) with an accessible visually-hidden pattern (`width: 1px; height: 1px; opacity: 0; clip: rect(0 0 0 0)`). This prevents the browser from ever scrolling to the "focused" hidden input.
- **Mirror the knob animation.** The LTR rule slides the knob from `right: 19px` (off) to `right: 0` (on). In RTL the switch is visually mirrored, so the knob now slides from `left: 19px` to `left: 0`.
## Notes

- Only `public/css/style-rtl.css` is changed; LTR behavior is untouched.
- The fix could also be applied to `style.css` so LTR benefits from the improved hiding technique (the focus-scroll issue technically exists in LTR too, just less visibly). Happy to do that in a follow-up PR if preferred.
## Testing

Reproduced on Chrome, Firefox, and Edge in Hebrew locale. After the fix, toggling any switch does not affect scroll position, and the knob animates in the correct direction.

### DIFF
--- a/public/css/style-rtl.css
+++ b/public/css/style-rtl.css
@@ -498,3 +498,32 @@ table.dataTable thead .sorting {
         margin-top: 0.5em !important;
     }
 }
+
+/**
+ Fix: In RTL, clicking the onoffswitch label moves focus to the hidden
+ checkbox positioned at left:-9999px, which causes the browser to scroll
+ the whole page to the left. Hide it without off-screen positioning.
+*/
+.onoffswitch-checkbox {
+        left: auto;
+        right: auto;
+        width: 1px;
+        height: 1px;
+        opacity: 0;
+        clip: rect(0 0 0 0);
+        overflow: hidden;
+        margin: 0;
+        padding: 0;
+        border: 0;
+}
+
+/* Mirror the knob: in LTR it slides from right:19px (off) to right:0 (on).
+   In RTL the label is visually mirrored, so anchor the knob to the left. */
+.onoffswitch-label:before {
+        left: 19px;
+        right: auto;
+}
+.onoffswitch-checkbox:checked + .onoffswitch-label:before {
+        left: 0;
+        right: auto;
+}


freescout-help:
Thanks. Pull requests should be sent to the `master` branch: https://github.com/freescout-help-desk/freescout/wiki/Development-Guide#github-workflow

Applied this way: https://github.com/freescout-help-desk/freescout/commit/17d31b62addae5de83185cca53aa7463fb7aa39c

PR #5399, opened 2026-05-08
Add configurable threshold to suppress transient fetch errors in Logs Monitoring

### LINKED ISSUE #5398: Add configurable threshold to suppress transient fetch errors in Logs Monitoring alerts
## **Problem**

The Logs Monitoring alert email (sent by freescout:logs-monitor) fires whenever any new fetch_errors row exists in the window, with no threshold or de-duplication. On many self-hosted installs, IMAP connections occasionally hiccup with errors like:

- Error: connection setup failed: Authentication failed; failed to authenticate
- Folder: INBOX; Error: failed to fetch messages
- Connection reset by peer / Connection timed out

These are typically transient (the next fetch run succeeds) but each one triggers a separate "Logs Monitoring" alert email. Admins end up with an inbox full of false-positive alerts and start ignoring, defeating the purpose of the alert.

**Today the only mitigations are:**

1. Untick "Fetch Errors" in alert_logs_names — silences real outages too.
2. Stretch alert_logs_period to day/week — delays detection of real outages.

Neither is a good answer.

## **Proposal**

Add an alert_logs_fetch_min_occurrences Option, exposed as a number input in Manage » Settings » Alerts inside the existing Logs Monitoring panel, labelled "Fetch Errors Threshold".

## **Behaviour:**

- LogsMonitor::handle() groups fetch_errors rows in the window by their error message (file/line stripped so the same error from different stack frames groups together).
- Rows whose message matches a known-transient pattern (connection setup failed, failed to fetch messages, failed to authenticate, Connection closed/reset/timed out/refused,
stream_socket_client, SSL operation failed, Temporary failure) are suppressed unless they recur ≥ N times within the window.
- Non-transient errors (auth misconfiguration, parse errors, etc.) and other log channels are unaffected.
- Default = 1, which preserves upstream behaviour exactly — fresh installs and existing installs see no change until an admin opts in.

cmer:
Resolves #5398.

## Summary

Adds an `alert_logs_fetch_min_occurrences` Option, exposed as a "Fetch Errors Threshold" number input in **Manage » Settings » Alerts** inside the existing Logs Monitoring panel. Lets admins suppress single-shot transient IMAP errors (auth races, brief connection drops) that recover on the next fetch run, while still escalating sustained failures.

**Default = 1**, which short-circuits the filter and matches upstream behaviour exactly — fresh installs and existing installs see no change until an admin opts in by raising the threshold.

## Behaviour

`LogsMonitor::handle()` now passes the fetched `ActivityLog` rows through `filterTransientFetchErrors()` before deciding whether to send the digest:

- Rows are grouped by error message, with the trailing `; File: …` stripped so the same error from different stack frames groups together.
- A row is dropped only if **both** conditions hold:
  - its message matches a known-transient substring (`connection setup failed`, `failed to fetch messages`, `failed to authenticate`, `Connection closed/reset/timed out/refused`, `stream_socket_client`, `SSL operation failed`, `Temporary failure`); **and**
  - the same fingerprint appears fewer than the configured threshold times in the window.
- Non-transient errors (mailbox config issues, parse errors, etc.) are never filtered.
- Other log channels (`send_errors`, `system`, …) are untouched.

When the threshold is `1` (default) the filter returns immediately, so there is no behaviour change and no measurable cost on existing installs.

## Files changed

- `app/Console/Commands/LogsMonitor.php` — filter helper + threshold lookup (+102 lines).
- `app/Http/Controllers/SettingsController.php` — read the new key in the `alerts` section's `Option::getOptions()` call.
- `resources/views/settings/alerts.blade.php` — number input + help text inside the existing `#alert_logs_wrap`. Both new strings wrapped in `__()`.

No DB migration, no route changes, no new dependencies.

## Test plan

- [ ] Fresh install (no Option set) — Logs Monitoring digest behaves identically to before.
- [ ] Set threshold to 1 via UI — same as default; verify single transient error still alerts.
- [ ] Set threshold to 3, simulate one transient `connection setup failed` row in the window — verify no alert email is sent and `freescout:logs-monitor` prints `Suppressed 1 transient fetch_errors entries (threshold 3)`.
- [ ] Set threshold to 3, simulate ≥3 identical transient rows — verify alert is sent and includes them.
- [ ] Verify a non-matching error (e.g. mailbox parse error) is never filtered regardless of count.
- [ ] Verify other channels (`send_errors`) are unaffected.

### DIFF
--- a/app/Console/Commands/LogsMonitor.php
+++ b/app/Console/Commands/LogsMonitor.php
@@ -21,16 +21,19 @@ class LogsMonitor extends Command
     protected $description = 'Send new log records by email';
 
     /**
-     * Minimum identical-error occurrences required for `fetch_errors`
-     * entries to be included in the digest. Single-shot transient
-     * IMAP blips fall below this and are dropped.
+     * Default minimum identical-error occurrences required for
+     * `fetch_errors` entries to be included in the digest. Used when
+     * the `alert_logs_fetch_min_occurrences` option is not set. The
+     * default is 1, which matches upstream behaviour (no filtering);
+     * raise it via the alerts settings page to suppress transient
+     * single-shot fetch errors.
      */
-    const FETCH_ERRORS_MIN_OCCURRENCES = 3;
+    const FETCH_ERRORS_MIN_OCCURRENCES_DEFAULT = 1;
 
     /**
      * Substrings that mark a `fetch_errors` row as transient. Rows
      * matching any of these are dropped regardless of count, unless
-     * they appear at least FETCH_ERRORS_MIN_OCCURRENCES times.
+     * they appear at least the configured min-occurrences times.
      */
     const FETCH_ERRORS_TRANSIENT_PATTERNS = [
         'connection setup failed',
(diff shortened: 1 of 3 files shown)


freescout-help:
Looks good. Just please make sure to avoid using references `&` in:
```
$filtered = $logs->reject(function ($log) use ($fetchLogName, $counts, $threshold, &$dropped)
```

freescout-help:
Added as single commit: https://github.com/freescout-help-desk/freescout/commit/a958553b723467a3a4338f766f61cb78000edbd7

PR #5428, opened 2026-05-24
Auto-reply: send from the alias the customer emailed

rosinghal:
## Summary

Currently the auto-reply is always sent from the mailbox's default email address — mailbox aliases and the "Allow to reply from aliases" (`aliases_reply`) setting are ignored on the auto-reply path. This PR makes the auto-reply behave the same way user replies already do:

- If the customer's email was sent to a mailbox alias **and** "Allow to reply from aliases" is enabled → the auto-reply is sent **from that alias** (with the configured alias display name where set).
- If `aliases_reply` is disabled, or the recipient is not one of the configured aliases → the auto-reply is sent from the **default mailbox email** (unchanged behaviour).

The implementation reuses the existing, proven alias pattern from `app/Mail/ReplyToCustomer.php` so behaviour is consistent across the app.

## Changes

- `app/Jobs/SendAutoReply.php` — match the original incoming thread's To/Cc against `Mailbox::getAliases(true, true)` (returns `[]` when `aliases_reply` is off, so the fallback to the default email is automatic) and pass the resolved alias to the `AutoReply` mailable.
- `app/Mail/AutoReply.php` — accept a `$from_alias` constructor argument and override the Swift `From` header inside the existing `withSwiftMessage` closure, mirroring the alias-handling block in `ReplyToCustomer::build()` (including From-Name precedence: alias display name only wins when the mailbox From Name is the plain mailbox name; a custom/user From Name still takes precedence).

No DB migration, config, or UI changes — `aliases`, `aliases_reply`, and the settings UI already exist.

## Test plan

- [ ] Mailbox with primary `support@x.com`, alias `sales@x.com(Sales Team)`, "Allow to reply from aliases" **enabled**, auto-reply enabled. Email `sales@x.com` → auto-reply `From:` is `Sales Team <sales@x.com>`.
- [ ] Same mailbox/alias, "Allow to reply from aliases" **disabled**. Email `sales@x.com` → auto-reply `From:` is the default `support@x.com`.
- [ ] With `aliases_reply` enabled, email an address that is **not** a configured alias → auto-reply `From:` is the default `support@x.com`.
- [ ] Email the default `support@x.com` directly → auto-reply `From:` is `support@x.com` (no regression).
- [ ] Set mailbox From Name to a custom value, repeat the alias case → custom From Name is used together with the alias address.
- [ ] `In-Reply-To`, `References`, and `Message-ID` headers are still set correctly on the auto-reply (no regression in `setHeaders`).

### DIFF
--- a/app/Jobs/SendAutoReply.php
+++ b/app/Jobs/SendAutoReply.php
@@ -77,9 +77,23 @@ class SendAutoReply implements ShouldQueue
         $failures = [];
         $exception = null;
 
+        // Reply from the alias the customer emailed (if "Allow to reply from aliases" is enabled).
+        $from_alias = '';
+        $aliases = $this->mailbox->getAliases(true, true);
+        if (count($aliases)) {
+            $original_recipients = array_merge($this->thread->getToArray(), $this->thread->getCcArray());
+            foreach ($original_recipients as $original_recipient) {
+                $original_recipient = \App\Email::sanitizeEmail($original_recipient);
+                if ($original_recipient && array_key_exists($original_recipient, $aliases)) {
+                    $from_alias = $original_recipient;
+                    break;
+                }
+            }
+        }
+
         try {
             Mail::to([['name' => $this->customer->getFullName(), 'email' => $customer_email]])
-                ->send(new AutoReply($this->conversation, $this->mailbox, $this->customer, $headers));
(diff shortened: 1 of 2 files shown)


freescout-help:

Applied as https://github.com/freescout-help-desk/freescout/commit/e2dcf6dc7e032ebf994b6adcb154f627b41dc4bf

Issue #4624, opened 2025-03-06
Problems fetching emails: 99 fetched instead of 100


fisthandle:
Hey,

i have repeating problems with emails fetching. I have a pro account on FastMail. When running fetch it often get a smaller number of emails than "page_limit", which breaks the query loop too soon. The result: it's not getting all emails that should be gotten.

Problem is with line [336](https://github.com/freescout-help-desk/freescout/blob/31c72b80928318a122dd73ff0b903e4beed83e3c/app/Console/Commands/FetchEmails.php#L336) of FetchEmails command:
`
} while (count($messages) == $page_size);`

So when the server gives 99 emails, not 100 (page_size), this stops after only one loop. It happens often enough that i'm missing some emails from last week. 

For now i fixed it with a workaround:
`} while (count($messages) > 10)`

I'd prefer to not change the main code. Can i ask for some help? Maybe i set something up in a wrong way? Has anyone had this problem before?
ChatGPT says that there may be some quirks of imap or maybe messages marked differently, which results in not all of them being sent. Maybe loop should be exited on a different condition?

fisthandle:
I've emptied my spam folder on FastMail, still it got 99, 99, 81 emails.

freescout-help:
Try Step 6: https://github.com/freescout-help-desk/freescout/wiki/Fetching-Emails#troubleshooting-fetching

fisthandle:
```
[2025-03-10 09:10:31] Fetching ALL emails for the last 1 days.
[2025-03-10 09:10:31] Folder: INBOX
[2025-03-10 09:11:04] Fetched: 97
(...)
[2025-03-10 09:11:56] Fetched: 97
(...)
[2025-03-10 09:12:19] Fetched: 0
[2025-03-10 09:12:20] Fetching finished
```

So it seems that i just not always gets a 100 messages. I don't know what can be the reason.

freescout-help:
We also don't know.

fisthandle:
OK, i'm not looking for a reason. But currently FS code for fetching emails is WRONG: it stops too quickly. I'd suggest fixing the condition for fetching loop. 

I'll keep my solution, but i guess you could consider modifying current code as it's not correct.

Issue #4047, opened 2024-05-29
Possible pagination conflict when fetching unseen messages


sofl88:
Hello.

Since we have some adjustments in our installation and some minor issues with the new fetching library, we did alot of testing recently.

I think I found a missconception in the fetchEmails command when it comes to pagination.
-> fetch over 300 (PAGE_SIZE), **unseen** messages.

When this line is executed the second time, the amount of returned emails changed because most of the messages were set to **seen** in the previous loop by **processMessage** > **setSeen**.
https://github.com/freescout-helpdesk/freescout/blob/2d871f27590424e2a33368878a01dd41ebb462e1/app/Console/Commands/FetchEmails.php#L223

So since the entire amount changes but the page is still increasing with every loop, some emails get ignored. I think it also depends on how fast the server flag these emails as seen.

It is not really an issue at the moment for us because we almost never fetch over 300 emails in one run but I thought it would be helpful to inform you guys.

PHP version: 8.0.30
FreeScout version: 1.8.140
Database: MySQL
Are you using CloudFlare: No
Are you using non-official modules: Yes

Thanks in advance

freescout-helpdesk:
Thanks for the info. We'll keep in mind. But it looks like so far no one had some issues due to this.

Issue #176, opened 2018-12-08
Error fetching Emails from Office365 (IMAP)


mdeuerlein:
IMAP connection to Office365 Account quits with error.

**Account settings**

Settings | Value
------------ | -------------
Protocol | IMAP
Server | outlook.office365.com
Port | 143
User | email@domain.com
Password | *****
Encryption | TLS

**Error Message**: `[BADCHARSET (US-ASCII)] The specified charset is not supported.`

I had the same Problem in a PHP Project but at the moment no time to contribute. But I think the following Links should help to solve the problem and if you need someone for testing, just let me know.

- **webklex/laravel-imap**
[Issue: Charset shouldn't be send with the imap_search](https://github.com/Webklex/laravel-imap/issues/100)
- **barbushin/php-imap**
[Issue: Problem with serverEncoding](https://github.com/barbushin/php-imap/issues/101#issuecomment-378136507)

crishnakh:
In this case I commented the charset parameter and the mail is fetching.
>freescout/vendor/webklex/laravel-imap/src/IMAP/Query/Query.php
Not the best solution.

Line 155
```php
/**
             * Don't set the charset if it isn't used - prevent strange outlook mail server errors
             * @see https://github.com/Webklex/laravel-imap/issues/100
             */
            if($this->getCharset() === null){
                $available_messages = imap_search($this->getClient()->getConnection(), $this->getRawQuery(), SE_UID);
            }else{
                $available_messages = imap_search($this->getClient()->getConnection(), $this->getRawQuery(), SE_UID); //, $this->getCharset()
            }
```

freescout-helpdesk:
https://medium.com/@freescout/oauth-2-0-g-suite-microsoft-365-and-php-7da16ca74314

phanousk:
Any news on this? It's almost a year so I wonder if there is any progress :)

freescout-helpdesk:
See https://github.com/freescout-helpdesk/freescout/wiki/OAuth-for-Google-and-Office-365

PR #5462, opened 2026-06-12
Fetching: paginate by SEARCH count so the newest emails are no longer dropped (#4624)

diiegopereira:
### Summary

On a busy mailbox, FreeScout can silently lose the **newest** incoming emails whenever a fetch window (`--days`) contains more messages than `fetching_bunch_size` **and** any non-final bunch fails to fully materialize. The dropped emails never appear and nothing actionable is logged. This is the problem reported in #4624 ("99 fetched instead of 100"), and is related to the pagination drift in #4047.

### Root cause

`app/Console/Commands/FetchEmails.php` drives its bunch loop off the **materialized** message count:

```php
$page = 0;
do {
    $messages_query->limit($page_size, $page); // webklex limit($limit, $page)
    $messages = $messages_query->get();
    // ... process ...
    $page++;
} while (count($messages) == $page_size);
```

Three interacting facts make this drop the newest mail:

1. **`while (count($messages) == $page_size)`** stops the loop the moment a bunch comes back with fewer than `$page_size` messages — but a bunch can come back short for reasons other than "end of mailbox" (a message that fails to materialize in the batch FETCH: malformed header, deleted between SEARCH and FETCH, etc.). One such message in an early bunch ends the loop **before the later bunches are ever fetched**.
2. **`fetch_order` defaults to `asc`** (oldest first), so the never-fetched later bunches are exactly the **newest** emails.
3. The loop starts at **`$page = 0`**, but webklex `Query::limit($limit, $page)` is **1-indexed** (`if ($page >= 1) $this->page = $page;`, default `1`) and `fetch()` uses `forPage($this->page, $this->limit)`. So `limit($size, 0)` is a no-op for the page and collapses to page 1 — the first bunch is fetched twice (harmless thanks to Message-ID dedup, but it confirms the off-by-one).

### Reproduction

- Mailbox with a 3-day window of ~106 messages, `fetching_bunch_size = 100`.
- One message in the first bunch fails to materialize in the batch FETCH → the bunch returns 99.
- `99 != 100` → the loop stops after the first bunch. The newest messages (the second bunch) are **never fetched** and silently lost.

### Fix

Compute how many bunches to fetch from a plain IMAP **SEARCH count** (no message bodies materialized), then iterate exactly `ceil(total / page_size)` 1-indexed pages. The loop is no longer driven by the fragile per-bunch materialized count:

```php
$total_messages = $folder->query()->since(...)->leaveUnread()->count(); // plain SEARCH
$total_pages    = ($page_size > 0) ? (int) ceil($total_messages / $page_size) : 1;
for ($page = 1; $page <= $total_pages; $page++) {
    // ... unchanged bunch body: limit($page_size, $page)->get(); process ...
}
```

### Why not `while (count($messages) > 0)`?

The obvious alternative — paginate until a page comes back empty — looks simpler but **introduces a crash**: terminating that way requires fetching one page *past* the end, and `forPage()` then returns an empty UID list. `ImapProtocol::fetch()` builds the message set as `$from . ':' . $from` with `$from = []`, throwing **`Array to string conversion`** on every fetch run. Counting up front never fetches an empty page, so it avoids both the data loss and this crash. (Alternatively, the empty UID list could be guarded inside `Query::fetch()` — happy to split that out if preferred.)

### Scope / caveats

- This fixes the loss for a **stable** search set (i.e. `fetch_unseen = 0`). With `fetch_unseen = 1`, `processMessage()` marks messages `\Seen` mid-loop, shrinking the unseen set between pages and shifting `forPage()` — the separate, pre-existing drift from #4047. That is not addressed here and is flagged in a code comment.
- The per-bunch charset fallback (#176) is preserved unchanged; the new `count()` mirrors the same `unseen` / `setCharset` conditions and falls back to a single bunch if the SEARCH itself fails.

### Testing

Validated read-only (IMAP PEEK, no writes) against a live high-volume mailbox: with the bunch size forced to 100 and 50, the old loop reaches 99/106 (drops the 7 newest); the count-based loop reaches 106/106 with no crash. An item-by-item probe confirmed all messages materialize individually (no intrinsically poisoned message — the short bunch is a transient batch-FETCH failure). Also exercised end-to-end with `fetching_bunch_size = 50` forcing 3 pages: all pages processed, no `Array to string conversion`.

Refs #4624, #4047

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -273,8 +273,43 @@ class FetchEmails extends Command
 
             // Requesting emails by bunches allows to fetch large amounts of emails
             // without problems with memory.
-            $page = 0;
-            do {
+            //
+            // Compute how many bunches to fetch from a plain IMAP SEARCH count up
+            // front, instead of driving the loop off the per-bunch materialized count.
+            // That avoids two bugs:
+            //  (1) Silent loss of the NEWEST emails (#4624): a single message that fails
+            //      to materialize in a batch fetch makes a bunch come back short; the old
+            //      `while (count($messages) == $page_size)` then stopped before fetching
+            //      the later bunches. With fetch_order='asc' those later bunches hold the
+            //      newest emails, so they were dropped silently.
+            //  (2) Looping "until an empty page" instead would fetch one page past the
+            //      end, and the IMAP layer throws "Array to string conversion" when asked
+            //      to fetch an empty UID list.
+            // Webklex Query::limit()/forPage() are 1-indexed, so pages start at 1.
+            //
+            // NOTE: this assumes a stable search set across pages (fetch_unseen=0). With
+            // fetch_unseen=1, processMessage() marks messages \Seen mid-loop, shrinking
+            // the unseen set between pages and shifting forPage() — a separate,
+            // pre-existing drift (#4047), not addressed here.
+            try {
+                $count_query = $folder->query()->since(now()->subDays($this->option('days')))->leaveUnread();
+                if ($unseen) {
+                    $count_query->unseen();
+                }
+                if ($no_charset) {
+                    $count_query->setCharset(null);
+                }
+                $total_messages = $count_query->count();
+            } catch (\Exception $e) {
+                // If the SEARCH count fails (e.g. charset issues on some servers), fall
+                // back to a single bunch so we never silently fetch nothing. The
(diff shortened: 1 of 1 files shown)


freescout-help:

Added as https://github.com/freescout-help-desk/freescout/commit/73a7e1f9dbbb5dcd229e9292666eea687cc7444e

PR #5471, opened 2026-06-19
feat: add Previous Message tab to Show Original modal

losimo:
# Add "Previous Message" tab to Show Original modal

## Problem

When an agent clicks **Show Original → Body** on a customer's reply, FreeScout shows only
the body of that specific message. The quoted content from the previous message is stripped
away — by design — to keep conversations clean.

However, this creates a blind spot: **a customer can silently modify the quoted text** when
replying, and the agent has no way to verify it without accessing the raw mailbox.

### Real-world example

A common support workflow sends a pre-filled form to the customer for validation:

> *"Here are your contract details. Please reply with 'OK' to confirm, or correct anything
> that is wrong. The bank account field is missing — please add it."*

The customer replies **"OK"**, but also fills in the bank account number (or corrects a
field) inside the quoted text. With the current UI, the agent only sees "OK". They must
leave FreeScout and open the raw mailbox to find the bank account the customer provided.

## Solution

Add a **"Previous Message"** tab to the Show Original modal. It renders the preceding
customer or agent message — the one being replied to — so agents can compare both messages
side by side without leaving FreeScout.

**Changes: 2 files, ~15 lines added.**

### `app/Http/Controllers/ConversationsController.php`

In `ajaxHtmlShowOriginal()`, fetch the immediately preceding `TYPE_CUSTOMER` or
`TYPE_MESSAGE` thread in the same conversation:

```php
$previous_thread = null;
if ($thread->conversation_id) {
    $previous_thread = Thread::where('conversation_id', $thread->conversation_id)
        ->where('id', '<', $thread->id)
        ->whereIn('type', [Thread::TYPE_CUSTOMER, Thread::TYPE_MESSAGE])
        ->whereNotNull('body')
        ->where('body', '!=', '')
        ->orderBy('id', 'desc')
        ->first();
}
```

Pass it to the view: `'previous_thread' => $previous_thread`.

### `resources/views/conversations/ajax_html/show_original.blade.php`

Add a conditional tab (only shown when a previous message exists):

```blade
@if ($previous_thread)
<li role="presentation">
    <a data-toggle="tab" href="#tab_previous_{{ $tabs_unique }}">{{ __('Previous Message') }}</a>
</li>
@endif
```

And the corresponding pane, rendering the body in a sandboxed iframe (consistent with the
existing Preview tab):

```blade
@if ($previous_thread)
<div id="tab_previous_{{ $tabs_unique }}" class="tab-pane fade">
    <div style="color:#888;font-size:12px;padding:6px 0 8px;border-bottom:1px solid #eee;margin-bottom:10px;">
        <strong>{{ $previous_thread->type == \App\Thread::TYPE_CUSTOMER ? __('Customer') : __('Agent') }}</strong>
        &mdash;
        {{ $previous_thread->created_at ? $previous_thread->created_at->format('d/m/Y H:i') : '' }}
    </div>
    <iframe sandbox=""
        srcdoc="{!! htmlspecialchars($previous_thread->getBodyOriginal(), ENT_QUOTES|ENT_SUBSTITUTE, 'UTF-8') !!}"
        frameborder="0"
        class="preview-iframe tab-body"></iframe>
</div>
@endif
```

## Security

- The iframe uses `sandbox=""` (same as the existing Preview tab), blocking scripts,
  forms, and cross-origin access.
- `htmlspecialchars(..., ENT_QUOTES|ENT_SUBSTITUTE)` correctly escapes the body for
  the `srcdoc` attribute context, preventing HTML injection through the attribute parser.

## Behaviour

- The tab only appears when there is a previous message in the same conversation
  (first message in a conversation: no tab shown).
- Only `TYPE_CUSTOMER` and `TYPE_MESSAGE` threads are considered — notes and line items
  are excluded.
- No new routes, no JavaScript, no extra HTTP requests. The previous message body is
  passed directly in the initial AJAX render.

## Translation

One new string: `'Previous Message'`. FreeScout uses the English string directly as the
translation key (no `en.json` exists), so English works automatically. Other locales should
add `"Previous Message": "<translation>"` to their respective JSON files under
`resources/lang/`.

### DIFF
--- a/app/Http/Controllers/ConversationsController.php
+++ b/app/Http/Controllers/ConversationsController.php
@@ -2561,10 +2561,22 @@ class ConversationsController extends Controller
             }
         }
 
+        $previous_thread = null;
+        if ($thread->conversation_id) {
+            $previous_thread = Thread::where('conversation_id', $thread->conversation_id)
+                ->where('id', '<', $thread->id)
+                ->whereIn('type', [Thread::TYPE_CUSTOMER, Thread::TYPE_MESSAGE])
+                ->whereNotNull('body')
+                ->where('body', '!=', '')
+                ->orderBy('id', 'desc')
+                ->first();
+        }
+
         return view('conversations/ajax_html/show_original', [
             'thread' => $thread,
             'body_preview' => $body_preview,
             'fetched' => $fetched,
+            'previous_thread' => $previous_thread,
         ]);
     }
 
--- a/resources/views/conversations/ajax_html/show_original.blade.php
+++ b/resources/views/conversations/ajax_html/show_original.blade.php
@@ -6,6 +6,7 @@
 	<li role="presentation" class="active"><a data-toggle="tab" href="#tab_preview_{{ $tabs_unique }}">{{ __('Preview') }}</a></li>
 	<li role="presentation"><a data-toggle="tab" href="#tab_body_{{ $tabs_unique }}">{{ __('Body') }}</a></li>
 	@if ($thread->headers)<li role="presentation"><a data-toggle="tab" href="#tab_headers_{{ $tabs_unique }}">{{ __('Email Headers') }}</a></li>@endif
+	@if ($previous_thread)<li role="presentation"><a data-toggle="tab" href="#tab_previous_{{ $tabs_unique }}">{{ __('Previous Message') }}</a></li>@endif
 </ul>
 
 <div class="tab-content">
@@ -21,4 +22,17 @@
 	@if ($thread->headers)<div id="tab_headers_{{ $tabs_unique }}" class="tab-pane fade">
 		<pre class="pre-wrap">{{ $thread->headers }}</pre>
 	</div>@endif
+	@if ($previous_thread)
+	<div id="tab_previous_{{ $tabs_unique }}" class="tab-pane fade">
+		<div style="color:#888;font-size:12px;padding:6px 0 8px;border-bottom:1px solid #eee;margin-bottom:10px;">
+			<strong>{{ $previous_thread->type == \App\Thread::TYPE_CUSTOMER ? __('Customer') : __('Agent') }}</strong>
+			&mdash;
+			{{ $previous_thread->created_at ? $previous_thread->created_at->format('d/m/Y H:i') : '' }}
+		</div>
+		<iframe sandbox=""
+			srcdoc="{!! htmlspecialchars($previous_thread->getBodyOriginal(), ENT_QUOTES|ENT_SUBSTITUTE, 'UTF-8') !!}"
+			frameborder="0"
+			class="preview-iframe tab-body"></iframe>
+	</div>
+	@endif
 </div>


freescout-help:
The "Show Original" window has Preview tab where full email history should be visible. FreeScout retrieves it via IMAP/POP3 from the mail server - if it's cropped, it means FreeScout could not retrieve it from the mail server and you need to figure out why.

Also customer's previous message is visible in the conversation history when viewing the conversation.

losimo:
Hi, thanks for the feedback — but I think there's been a misunderstanding about what this PR does.

  **This PR is not about the "Preview" tab or email retrieval from the mail server**: What it adds is a new "_Previous Message_" tab inside the "_Show Original_" modal. This tab displays the previous customer message from the conversation thread already stored in FreeScout's database — no IMAP/POP3 involved.

  **The use case**: An agent opens "_Show Original_" on a customer reply to inspect its raw HTML (e.g. to check for hidden formatting, malicious content, or modified quoted text like an IBAN added in a contract validation reply). At that point they are focused on that modal — switching back to the conversation timeline to read the prior message breaks their workflow.

The proposed tab lets them see the previous message's content **without leaving the modal**.

**Regarding "customer's previous message is visible in the conversation history"**: Yes, it is — but the point is to avoid context-switching when you're already inside the Show Original modal inspecting raw content.

The implementation queries **Thread** records already in the database (no external calls), renders the previous body in a sandboxed iframe, and adds ~26 lines across 2 files.

  Happy to clarify further or adjust the implementation if there are concerns about the approach itself.

freescout-help:
The thing is that over all the years this is the first time this feature is requested. Also this new tab may lead to confusion: the Body tab shows the message retrieved via IMAP/POP3 and the new tab will be showing content from DB.

So it would be better to add it via a custom module.

losimo:
Thanks — I understand the concern about keeping the UI unambiguous.

That said, I don’t think “this is the first time it’s been requested” is, by itself, a strong reason to dismiss the issue. In this case the problem is still real: when a customer edits the quoted text in a reply, the agent may not notice it from within the current inspection flow, even though that information can materially change the meaning of the message.

I also think the confusion concern can be addressed quite clearly in the UI. The proposed tab is not meant to represent the same data source as the Body tab. The Body tab continues to show the original message body as retrieved for that message, while the new tab explicitly shows the previous message from the conversation history already stored in FreeScout. They serve different purposes and can be labeled accordingly, so I don’t see that distinction as inherently confusing.

The practical value of the tab is that it avoids breaking the agent’s workflow. When someone opens Show Original, they are already in an inspection context. Asking them to leave the modal, scan the conversation timeline, and then mentally compare content across views adds friction to a task that should be straightforward.

So the goal of this PR is not to replace existing conversation history, nor to change how Preview/Body work. It is simply to make the relevant previous message available inside the same inspection modal, where the agent is already working.

If the main objection is specifically that the data comes from two different sources, I’d be happy to adjust the wording in the UI to make that explicit — for example by renaming the tab to something more precise, such as “Previous Conversation Message”, and/or adding a short note indicating that it is loaded from the conversation record rather than the raw email source.

I understand the suggestion to keep this as a custom module, but from my perspective this solves a general usability gap in the core product rather than a niche integration-specific need.

PR #5474, opened 2026-06-23
fix: prevent sending blank email when Summernote leaves empty HTML tags

### LINKED ISSUE #4590: Bug: You can send a blank email if you enter text and remove it
I was able to reproduce this on the live demo site.

If you make a new message (or a reply) with no content and press 'send' it properly tells you that you need to enter some text.

However if you enter text and _delete_ it, you can send the email without content.

losimo:
## Problem

Closes #4590.

When a user types text in the reply editor (Summernote) and then **deletes it all**, the underlying `<textarea id="body">` is not truly empty — Summernote leaves residual HTML like `<p><br></p>` or `<div><br></div>`. Parsley's built-in `required` validator sees a non-empty string and lets the form submit, resulting in a blank email being sent.

Steps to reproduce (also works on the live demo):
1. Open a conversation reply editor
2. Type some text
3. Select all and delete it
4. Click **Send** — the email is sent with no content

## Root cause

The `<textarea>` uses `data-parsley-required="true"`, which checks `value !== ''`. After type-and-delete, Summernote leaves `<p><br></p>` in the textarea — non-empty string, validation passes.

## Fix

Add a custom Parsley validator `htmlRequired` that strips HTML tags and `&nbsp;` entities before checking for real text content:

```js
window.Parsley.addValidator('htmlRequired', {
    validateString: function(value) {
        return !!$.trim(value.replace(/<[^>]+>/g, '').replace(/&nbsp;/gi, ''));
    },
    messages: { en: 'This field is required.' }
});
```

Replace `data-parsley-required` with `data-parsley-html-required` on the `body` textarea in both `view.blade.php` and `create.blade.php`.

## Tested cases

| Input | Result |
|-------|--------|
| `<p><br></p>` (Summernote empty) | ❌ blocked |
| `<div><br></div>` (Summernote empty) | ❌ blocked |
| `<p>&nbsp;</p>` (only non-breaking space) | ❌ blocked |
| `""` (truly empty) | ❌ blocked |
| `<p>Hello</p>` | ✅ allowed |
| `<p><strong>Text</strong></p>` | ✅ allowed |

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -90,6 +90,15 @@ $.extend(window.ParsleyConfig, {
     errorTemplate: '<div></div>'
 });
 
+// Custom validator for HTML editor fields (Summernote leaves <p><br></p> when emptied).
+// Used via data-parsley-html-required on the body textarea.
+window.Parsley.addValidator('htmlRequired', {
+    validateString: function(value) {
+        return !!$.trim(value.replace(/<[^>]+>/g, '').replace(/&nbsp;/gi, ''));
+    },
+    messages: { en: 'This field is required.' }
+});
+
 // Push notifications
 /*Push.config({
     serviceWorker: './customServiceWorker.js', // Sets a custom service worker script
@@ -1659,12 +1668,8 @@ function showReplyForm(data, scroll_offset)
 		$('#body').summernote('focus');
 	}
 
-	if (!isChatMode()) {
-		// Select2 for CC/BCC
-		initRecipientSelector();
-	} else {
-		$('form.form-reply:first .field-cc').addClass('hidden');
-	}
+	// Select2 for CC/BCC
+	initRecipientSelector();
 
 	if (typeof(scroll_offset) == "undefined") {
 		scroll_offset = 0;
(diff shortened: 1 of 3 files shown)


freescout-help:
1) Make sure to use the latest version of FreeScout.

2) Now the send for allows to send `<div><br></div>` reply (just open reply form and click Send).

3) The `data-parsley-required-message` parameter is not respected.

losimo:
Thanks for the review! I've addressed all three points and opened a clean replacement PR: #5478

**Changes made:**
1. Rebased on latest upstream `dist` — the new PR is based on the current HEAD
2. Fixed the blank-send bug: instead of a custom Parsley validator, the fix now clears the textarea to `''` just before `parsley().validate()` is called, so the standard `data-parsley-required` catches visually-empty Summernote content (`<div><br></div>`, etc.)
3. No blade template changes — `data-parsley-required` and `data-parsley-required-message` are used as-is from upstream, so the error message is respected correctly

The fix is a single clean commit touching only `public/js/main.js`.

PR #5475, opened 2026-06-23
fix: Customer Waiting Since shows time of last customer message instead of first

### LINKED ISSUE #5225: Incorrect Customer Waiting Since metric
Hello,

I noticed Customer Waiting Since metric is a bit incorrect.

For instance:
* customer's message 1 posted at 10:00
* customer's message 2 posted at 10:30
* no user messages between these two messages

In the case customer is waiting since 10:00, not 10:30. But customer waiting since is set to 10:30.

losimo:
## Problem

Closes #5225.

When a customer sends multiple messages without receiving a staff reply, the **Customer Waiting Since** metric advances to the time of each new message instead of staying at the first unanswered message.

**Example:**
- Customer sends message 1 at 10:00
- Customer sends message 2 at 10:30 (no staff reply in between)
- Metric shows: *waiting since 10:30* ❌
- Expected: *waiting since 10:00* ✅

## Root cause

`last_reply_at` — the field used by `getWaitingSinceField()` for active conversations — is unconditionally updated to `$now` on every incoming customer message in two places:

- `app/Console/Commands/FetchEmails.php:1074` (incoming email via IMAP)
- `app/Observers/ThreadObserver.php:46` (thread creation observer, covers API and web)

Both update `last_reply_at` regardless of whether `last_reply_from` is already `PERSON_CUSTOMER`, overwriting the original waiting time with each new customer message.

## Fix

Skip the `last_reply_at` update when the incoming thread is from a customer **and** the previous `last_reply_from` was already `PERSON_CUSTOMER`. Staff replies always reset the timer correctly.

```php
// Only update last_reply_at if the customer was not already the last to reply.
if ($conversation->last_reply_from != Conversation::PERSON_CUSTOMER) {
    $conversation->last_reply_at = $now;
}
```

Applied identically in both `FetchEmails.php` and `ThreadObserver.php`.

## Tested scenarios

| Scenario | Expected `last_reply_at` | Result |
|----------|--------------------------|--------|
| Two consecutive customer messages | Time of first message | ✅ |
| Three consecutive customer messages | Time of first message | ✅ |
| Customer → staff → customer | Time of last customer message | ✅ |
| First message (new conversation) | Message time | ✅ |
| Two consecutive staff messages | Time of last staff message | ✅ |


freescout-help:
Added as https://github.com/freescout-help-desk/freescout/commit/caa5aac1c3ae4602faad7d9332f7e19664f23852

freescout-help:
The update has been reverted as it broke conversations order: https://github.com/freescout-help-desk/freescout/issues/5501#issuecomment-5084418616

PR #5476, opened 2026-06-23
fix: CMD+Enter does not send reply on Mac

### LINKED ISSUE #4425: Keyboard Shortcut Module - Command-Enter Not working on Mac
Keyboard Shortcut Module is installed.

On a macbook, command-enter is not working to send the message.  I have tried on Firefox and Chrome.

PHP version: 8.3.14
FreeScout version:  1.8.159
Database: PostgreSQL
Are you using CloudFlare: No
Are you using non-official modules:  No

losimo:
## Problem

Closes #4425.

On macOS, **CMD+Enter** does not send the reply. Ctrl+Enter works (via the KeyboardShortcuts module), but Mac users naturally expect CMD+Enter — the same shortcut used by Gmail, Slack, Apple Mail, etc.

## Root cause

Two issues combine to produce this:

1. The chat-mode `keydown` handler in `main.js` explicitly returns early when `e.metaKey` is true (line 1418), so CMD+Enter is intentionally ignored there.
2. The KeyboardShortcuts module only checks `e.ctrlKey` for its send shortcut, not `e.metaKey`. On Mac, the Command key sets `metaKey`, not `ctrlKey`, so the module's handler never fires for CMD+Enter.

Evidence from issue comments — debug output when pressing CMD+Enter on Mac:
```
keydown {key: 'Enter', code: 'Enter', which: 13, ctrlKey: false, metaKey: true}
// module handler fires with metaKey: false — doesn't match, event ignored
```

## Fix

Add a dedicated `keydown` handler in `initReplyForm()` that triggers on `metaKey + Enter`, independently of the KeyboardShortcuts module:

```js
$(document).on('keydown.cmd-enter-send', function(e) {
    if (!e.metaKey || e.which != 13 || e.altKey || e.shiftKey) return;
    if (isChatMode() || $('.modal:visible').length) return;
    if (!$(':focus').hasClass('note-editable')) return;
    var button = $('div.conv-reply-body:visible .btn-reply-submit:first');
    if (button.length) {
        e.preventDefault();
        button.click();
    }
});
```

**Why in the core and not the module:** CMD+Enter is the standard Mac send shortcut across all major apps. It belongs in the core for all Mac users, not gated behind a module. The `fs_processing_send_reply` guard already prevents double-sends if the module also fires.


freescout-help:
Applied as https://github.com/freescout-help-desk/freescout/commit/460bc8f91b88a5400698f1a8eae141d413edc542

PR #5477, opened 2026-06-23
fix: strip trailing &gt;/&lt; HTML entity from linkified URLs

### LINKED ISSUE #5423: wrong tag in purified URL
In some cases URLs with passwords including special characters are submitted such as here:

Image

Unfortunately the trailing `>` is being processed as a part of the URL in this case which leads to a negative user experience, because the submitted password isn't working.

PHP version: 8.5.5
FreeScout version: 1.8.220
Database: MySQL
Are you using CloudFlare: No

losimo:
## Problem

Closes #5423.

When a URL in a plain-text email body ends with `>` (e.g. a password containing `>`), `purifyHtml()` correctly encodes `>` as `&gt;` in the HTML text node. However, `linkify()` then includes the HTML entity `&gt;` as part of the URL in the generated `href`, because `[^\s()<>]+` treats `&`, `g`, `t`, `;` as ordinary URL characters — it only excludes the literal `>` character, not its entity form.

The regex trailing-char check (`[^[:punct:]\s]`) strips the final `;` (punctuation) but leaves `&gt` (without `;`) in the captured URL. This incomplete entity gets placed inside the `href` attribute:

```html
<!-- Input text after purifyHtml() -->
Reset password: https://example.com/?token=abc&gt;

<!-- Output of linkify() — BUG -->
Reset password: <a href="https://example.com/?token=abc&gt">...</a>;
```

HTML5 browsers decode `&gt` (without `;`) as `>`, so the link navigates to `https://example.com/?token=abc>` — a URL with a stray `>` that the server rejects or mishandles.

## Root cause

`linkify()` in `Helper::linkify()` uses `[^\s()<>]+` for the URL body, which excludes the literal `>` character but not the multi-character HTML entity `&gt;`. After `purifyHtml()` runs, `>` in text content is always encoded as `&gt;`, so the literal `>` exclusion no longer provides protection.

## Fix

Strip trailing `&gt;` and `&lt;` entities (with or without trailing `;`) from the captured URL before building the `href`:

```php
// Strip trailing HTML angle-bracket entities (&gt; or &lt;, with or without
// semicolon) that purifyHtml() introduces when > or < appear at the end of
// a plain-text URL (issue #5423).
$link = preg_replace('/&(?:gt|lt);?$/', '', $link);
```

Applied in the `http`/`https` callback inside `linkify()`, before `substr()` removes the scheme prefix.

## Test

| Input | href before fix | href after fix |
|-------|----------------|----------------|
| `https://example.com/?token=abc&gt;` | `...abc&gt` ❌ | `...abc` ✅ |
| `https://example.com/normal` | `...normal` ✅ | `...normal` ✅ |
| `https://example.com/path/` | `.../path/` ✅ | `.../path/` ✅ |


freescout-help:
Unfortunately the fix does not work.

Try to send a reply on behalf of a support agent with the following text:
```
https://example.com/?token=abc>
```

It will look like:

losimo:
Thanks for testing and for the clear reproduction case.

You're right that the previous approach (stripping `&gt` from the captured URL with `preg_replace` after the match) was incomplete — even though the `href` was fixed, the regex was still capturing `&gt` up to the trailing `;` exclusion, leaving an orphaned `;` as visible text after the link.

I've reworked the fix to address the root cause at the **regex level** instead:

**Old pattern** (in the URL body):
```
[^\s()<>]+
```

**New pattern** (tempered greedy token):
```
(?:(?!&(?:gt|lt);)[^\s()<>])+
```

The negative lookahead `(?!&(?:gt|lt);)` prevents the URL body from consuming an HTML angle-bracket entity, so `&gt;` is never included in the captured URL in the first place. The entity remains as text after the link, where the browser decodes it as `>`.

Result for `https://example.com/?token=abc>`:
| | `href` | Link text | Visible |
|---|---|---|---|
| Before (no fix) | `…abc&gt` | `…abc&gt` → `abc>` | `abc>;` |
| After (regex fix) | `…abc` | `…abc` | `abc>` |

Also removed the now-unnecessary `preg_replace` post-processing step.

freescout-help:
Applied as https://github.com/freescout-help-desk/freescout/commit/30ba029242a2a35fa5bfa245123627ad332ebb22

PR #5493, opened 2026-07-07
Append to logs instead of overwritten

neurit-matejka:
Log files of `fetch-emails.log` and `queue-jobs.log` are **overwritten** with each call. I think, it should be appended.
This pull request changes used Laravel's functions `sendOutputTo` to `appendOutputTo`.


freescout-help:
It would be better to allow to enable appending via .env file variable.

PR #5497, opened 2026-07-13
Add another Docker image's link to README.md

luixxiul:
Hello! https://github.com/etkecc/freescout provides a hardened Docker Image of FreeScout, which runs as non-root user with all Linux capabilities dropped on read-only rootfs, and I think it should be deserve to be mentioned on README.


freescout-help:
Thanks. We'll see how this image evolves and maybe add it in the future.

PR #5517, opened 2026-07-21
Prevent forwarded notifications from replying to customers

### LINKED ISSUE #4515: Forwarded email notifications in FreeScout incorrectly linked to original conversation and notify external sender
When a user forwards an email from one FreeScout mailbox to another, the system incorrectly associates the forwarded email with the original conversation. Additionally, the original sender is notified of this forwarded email, which is unintended and could cause confusion or confidentiality issues.

**Steps to Reproduce:**

- Receive an email from an external sender in a FreeScout mailbox.
- Assign the email to a user within FreeScout.
- The assigned user replies to the email notification, forwarding it to another FreeScout mailbox.
- Observe how the forwarded email is handled by the second mailbox and whether the original sender is notified.

**Expected Behavior:**

- The forwarded email should either:
- Be treated as a new thread in the new mailbox, or
- Be clearly marked as forwarded while retaining context, without notifying the original sender.

**Actual Behavior:**

- FreeScout assigns the forwarded email to the original conversation instead of treating it as a new thread or marking it distinctly as forwarded.
- The original sender is also notified of the forwarded email, which is unintended and problematic.

**Impact:**

- Misclassification of emails disrupts workflows and creates confusion in shared inboxes.
- Notifying the original sender of internal forwarding breaches confidentiality and could lead to misunderstandings.

**Suggested Solution:**

- Implement logic to differentiate forwarded emails from replies to the original thread.
- Ensure forwarded emails are treated as new threads or clearly marked as forwarded.
- Prevent notifications to the original sender for emails forwarded internally between FreeScout mailboxes.

patattzel:
Fixes #4515.

## Summary

When a support agent forwards a FreeScout user notification to another FreeScout mailbox, mail clients may preserve the notification Message-ID in `In-Reply-To`, `References`, or the hidden body marker. FreeScout then treats the forward as an agent reply and may send the internal text to the original customer.

This change detects forwarded user notifications before conversation threading and starts a new conversation instead of attaching the message to the original conversation.

## Safety

- Applies only when the subject has a recognized forward prefix and a valid `FS_notify` Message-ID is present.
- Validates the notification hash using `Mail::getMessageIdHash()`.
- Does not change normal notification replies or `FS_reply` handling.
- Does not inspect arbitrary body text for forwarding markers, avoiding false positives from quoted conversation history.
- Supports a second FreeScout mailbox in CC.

Email has no standardized forward header, so subject-prefix detection is intentionally a pragmatic heuristic. If a sender removes or replaces the forward prefix with an unknown prefix, the forward cannot be distinguished reliably from a reply.

## Tests

Added unit and integration coverage for:

- Forwarded notification to Mailbox B in CC creates a new conversation.
- No `SendReplyToCustomer` job is dispatched for the original customer.
- Hidden FreeScout marker fallback when reply headers are removed.
- Normal notification replies retain existing behavior.
- Replies still work after a conversation is moved between mailboxes.
- Invalid notification hashes and non-notification Message-IDs are ignored.

```
OK (32 tests, 97 assertions)
```

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -552,7 +552,7 @@ class FetchEmails extends Command
             // in In-Reply-To/References or in the hidden marker. Do not attach such
             // forwards to the original conversation, otherwise the original
             // customer may receive the forwarded internal message.
-            if ($this->shouldStartNewConversationForForward($subject, $prev_message_ids)) {
+            if ($this->shouldStartNewConversationForForward($subject, $prev_message_ids, $in_reply_to)) {
                 $this->line('['.date('Y-m-d H:i:s').'] Forwarded FreeScout notification detected. Creating a new conversation.');
                 $prev_message_ids = [];
             }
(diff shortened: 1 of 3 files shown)


freescout-help:
This update breaks existing logic in the following case:

1) Customer sends an email with `Fwd: Test` subject.
2) Support agent replies to the received email notification.
3) New conversation will be created instead of adding support agent's reply to the existing conversation.

patattzel:
Thank you — this was a valid false-positive. I have addressed it in commit 6b062aa0.

The detection now distinguishes an existing `Fwd:` in the customer conversation subject from a forward prefix newly added to the notification subject:

- A valid notification `In-Reply-To` is treated as an authoritative reply signal.
- Without `In-Reply-To`, the subject after removing the outer forward prefix (and the default `[#number]` notification prefix) must match the original conversation subject.

Regression coverage now includes both sides of this case:

1. Customer subject `Fwd: Test` + agent reply with subject `Fwd: Test` -> appended to the original conversation and reply queued to the customer.
2. Customer subject `Fwd: Test` + agent actually forwards the notification (`Fwd: [#number] Fwd: Test`) -> new conversation and no customer reply job.

Full test suite: `OK (34 tests, 101 assertions)`.

freescout-help:
Both forwarded and simply replied by user agent email notifications have `In-Reply-To` header properly set up. So checking it is no use.

Here is the proper regex for checking FWD prefix (it's much broader than you think): https://github.com/freescout-help-desk/freescout/blob/dist/app/Console/Commands/FetchEmails.php#L855

So at the moment this solution still can't properly distinguish forwarded email notifications and regular replies to email notifications (for emails containing something like "Fwd:" in their subject).

patattzel:
Thanks, you are right. I underestimated the scope here. With the current notification format, there is no reliable generic way to distinguish replies from forwards.

I can narrow the fix to the cross-mailbox safety issue from #4515 by comparing the mailbox that sent the notification with the mailbox receiving the message. Regular replies would remain unchanged, including conversations with `Fwd:` in the original subject. The existing broad subject-prefix regex would be used, and `In-Reply-To` would no longer be treated as a distinguishing signal.

Would you consider this narrower safeguard acceptable, or would you prefer the source mailbox to be encoded in the notification Message-ID or hidden marker?

Thanks again for the careful review. It helped uncover assumptions in the first implementation.

freescout-help:
Fix applied: https://github.com/freescout-help-desk/freescout/commit/9d16a0d4d306264425a87fa7d929bc9af5f75615

Issue #123, opened 2018-11-13
Support for Amazon SES


guestisp:
Amazon SES will replace any message-id header with something custom.

We can't rely on message-id to track messages in each conversation and currently sending through Amazon SES will break freescout

Something different should be used

freescout-helpdesk:
Amazon SES moves `Message-ID` into `In-Reply-To` / `References`. FreeScout analyses these headers, so app need to be tested with Amazon SES and modified if needed to properly fetch Message-ID from `In-Reply-To` / `References`.

guestisp:
We are using Amazon SES and we were unable to fix this.
Hoping someone else could fix :)

freescout-helpdesk:
Working on this

guestisp:
Thank you, so we can use Amazon SES also for freescout.

PR #5518, opened 2026-07-21
Rename "Note" to "Internal Note" throughout the agent UI

adesegun-threls:
The client's source system already treats "Note" as internal by convention, so plain "Note" next to a customer reply here reads as ambiguous, is it an internal note or something else. This renames it to "Internal Note" everywhere it actually means that, using the same translation-override technique already used for Closed to Solved and Active to New.

"Note" turned out to be just as overloaded a word as "Active" was. Customer::$notes is a real, unrelated customer field, and one of our own ArmsReports placeholder columns reused the exact same key too. I left both alone and gave the ArmsReports placeholder its own "Details" header instead of touching the shared key.

Everywhere else that genuinely means the internal-note feature got its own translation key: the reply/note toggle button, the "Add Note" submit button, the print-view tag, the warning shown when you're about to reply to someone else's note, the activity line under each note thread ("X added a note to conversation #123"), both note-added flash messages, the default-redirect settings text, the subscription notification copy, and four labels inside Workflows. None of those 15 keys share anything with the two collisions, and none collide with each other either, so unlike Active to New this didn't need a runtime patch command anywhere, even to reach the paid Workflows module.

I tested that every one of those strings resolves to its new wording, that Customer::$notes and the ArmsReports placeholder still read exactly as before, and that the underlying Thread::TYPE_NOTE value is untouched.


adesegun-threls:
Opened against the wrong repository by mistake, this was meant for a private fork, not upstream. Closing.

PR #5519, opened 2026-07-22
Auto-focus the reply editor when opening a ticket

adesegun-threls:
Client feedback (ARMS-43): agents have to click Reply before they can start typing on every ticket, even though most tickets get a reply, not a note. That's an extra click on every single ticket, and it adds up across the kind of daily volume ARMS handles.

Turns out chat mode already does this. It auto-opens and focuses the reply editor on load, it's just gated to chat conversations only. I extended that same behavior to regular email tickets, reusing the exact same open/focus code path chat mode already uses rather than building something new.

I kept the existing priority order intact: if there's a cached note draft sitting in the browser for that ticket, or a draft actively being restored, that wins and the editor doesn't get force-switched back to Reply under it. Only a completely fresh ticket, with nothing else going on, gets auto-opened to Reply.

I checked that opening the editor this way has no side effects of its own (no status change, no ajax call, nothing happens until the agent actually hits Send), and traced through the phone/chat-only edge cases where the Reply button doesn't even exist in the page, which is a safe no-op.

Still worth a manual pass before this goes out: how it behaves on narrow/mobile screens (auto-focus pops the keyboard), and Closed/Spam conversations specifically, just to eyeball that nothing unexpected happens there.


adesegun-threls:
Opened in error against the upstream repo instead of our fork. Closing immediately, no action needed here.

freescout-help:
Thanks. But we don't plan to show automatically the reply form on opening a conversation.

PR #5521, opened 2026-07-23
Make Shebang in update.sh generic

utrenkner:
The more generic Shebang works also on e.g. FreeBSD, which has the Bash executable at `/usr/local/bin/bash`.


freescout-help:
Thanks. Applied as https://github.com/freescout-help-desk/freescout/commit/e6721a094b21fa1e869ce055a7efb30ea2c96b57

PR #5522, opened 2026-07-23
fix: linkify plain URLs in the outgoing customer reply email

### LINKED ISSUE #5280: Plain URLs typed/pasted in editor are not auto-linked in outgoing emails
PHP version: 7.4.33
FreeScout version: 1.8.209
Database: MySQL
Are you using CloudFlare: No

## Description

When typing or pasting a plain URL into the reply editor and sending the message, 
the URL arrives as plain text in the customer's email — not as a clickable link.

## Steps to Reproduce

1. Open a conversation and click "Reply"
2. Type or paste a bare URL into the editor, e.g. `https://example.com`
3. Click "Send"
4. Check the received email source

**Result:** The URL is sent as plain text — not clickable.

## Expected Behavior

Plain URLs should be converted to clickable links before sending.

## Additional observation

Summernote already has auto-link capability — when typing a URL and pressing 
**Enter**, it converts the URL into a clickable link. However this only triggers 
on Enter, not reliably in all cases. A server-side auto-link before sending 
(similar to what was done for incoming emails in #341) would be more robust.

## Raw email source (excerpt)

Sent as:
<div>https://example.com</div>

Expected:
<div><a href="https://example.com">https://example.com</a></div>

raphyabak:
## Summary

The customer-facing reply email rendered the thread body via `getCleanBody()` (sanitize only), while the web UI's conversation view uses `getBodyWithFormatedLinks()` (sanitize + `linkify()` to convert plain-text URLs into real `<a>` tags). So a plain URL an agent typed looked correctly clickable in FreeScout's own conversation view, but arrived as non-clickable plain text in the customer's inbox. Matches #5280.

- `getBodyWithFormatedLinks()` already exists, already calls `getCleanBody()` internally, and is already used elsewhere for display — this just points the outgoing customer email template at it instead of the more limited `getCleanBody()`.

Fixes #5280

This also covers the second case from the report (a link Summernote auto-creates when typing a URL and pressing Enter also arriving non-clickable): I verified `getCleanBody()` alone already preserves a real `<a href>` tag untouched (HTMLPurifier's config explicitly allows `a[href]`), so that symptom can only occur if the URL is still plain text by the time it reaches the server — which `linkify()` now catches regardless of why it stayed plain text. `getBodyWithFormatedLinks()` also leaves an already-real `<a>` tag alone rather than double-wrapping it, so both cases resolve through the same fix.

## Test plan

- [x] Added `tests/Unit/ThreadBodyLinkifyTest.php`:
  - confirms `getCleanBody()` does not linkify plain URLs
  - confirms `getBodyWithFormatedLinks()` does linkify plain URLs
  - confirms it preserves pre-existing `<a>` tags without duplicating them
  - statically guards that the customer reply template calls `getBodyWithFormatedLinks()`, not `getCleanBody()`
- [x] Verified each new test fails without the fix and passes with it.
- [x] Full test suite passes (29 tests, 78 assertions), no regressions.


freescout-help:
As mentioned in https://github.com/freescout-help-desk/freescout/issues/5280:

> Plain URLs should be converted to clickable links before sending.

Your solution will have side effects - it may convert into links URLs which support agent does not want to be converted.

freescout-help:
https://github.com/freescout-help-desk/freescout/issues/5280#issuecomment-5067602711

Issue #5222, opened 2026-02-18
Workflow: Trigger when customer replies


jonalange:
As stated on the workflow page I used the `New / Reply / Moved` as trigger.

<img width="749" height="120" alt="Image" src="https://github.com/user-attachments/assets/135d673f-477f-4433-a563-9fd7758d6ece" />

But it doesn't seem to trigger when the customer replies. 

Is this because I set `Max Executions` to `1`
So 1 is across the whole ticket, not per individual reply? 

<img width="1315" height="442" alt="Image" src="https://github.com/user-attachments/assets/5481de0e-32c8-42e0-81a4-2cb02c6c85b0" />

What would be the correct way of handling this? I want to add a tag each time a customer replies as I use that as a filter for the API.
Should I just set `Max Executions` to something very high like `999` or what would the correct way of handling this?
Or do I need to set `Apply to previous` to true as it's technically an already existing conversation?

Also, this would also trigger if we reply to the customer, right? There is no true "only when customer replies" trigger?

freescout-help:
> So 1 is across the whole ticket, not per individual reply?

Yes, per ticket.

> Or do I need to set Apply to previous to true as it's technically an already existing conversation?

Yes.

jonalange:
But this would also trigger if we reply to the customer, right? There is no true "only when customer replies" trigger?
If not this would be a good addition as trigger.

MarkoD:
@jonalange You can add a 2nd condition "Customer Waiting" which triggers only when a customer replies. Combining these conditions you can simulate a "Customer (not agent) replied" trigger.

Note: The "Customer Waiting" condition is reset each time a customer replies, which may be considered a bug (https://github.com/freescout-help-desk/freescout/issues/5225). Currently, however, this suits your needs.

jonalange:
> [@jonalange](https://github.com/jonalange) You can add a 2nd condition "Customer Waiting" which triggers only when a customer replies. Combining these conditions you can simulate a "Customer (not agent) replied" trigger.
> 
> Note: The "Customer Waiting" condition is reset each time a customer replies, which may be considered a bug ([#5225](https://github.com/freescout-help-desk/freescout/issues/5225)). Currently, however, this suits your needs.

Great Idea, thank you very much!
Will try that.


But it would still be nice to have this natively anyway.

MarkoD:
[...]

PR #5528, opened 2026-07-26
fix: compute Customer Waiting Since from thread history instead of last_reply_at

### LINKED ISSUE #5225: Incorrect Customer Waiting Since metric

### LINKED ISSUE #5501: The conversation doesn't appear in the list when the customer replies several times
A customer sent an email to an existing conversation today, but the conversation got buried somewhere at the bottom of the list and didn't make it to the top.

Image

Image

PHP version: 8.4.23
FreeScout version: 1.8.229
Database: MySQL
Are you using CloudFlare: No

losimo:
## Problem

PR #5475 fixed #5225 (*Customer Waiting Since* jumping to the time of each new customer message instead of staying at the first unanswered one) by skipping the `last_reply_at` update whenever the customer was already the last to reply.

That broke conversation ordering (#5501): `last_reply_at` is also the field `Folder::getOrderByArray()` sorts folder lists by (`desc`), so freezing it on repeat customer messages stopped those conversations from bubbling to the top — a customer could send a 2nd or 3rd follow-up and the conversation would stay stuck in its old position, making it easy for agents to miss. The fix was reverted.

## Root cause

`last_reply_at` is overloaded with two incompatible responsibilities:
1. **Recency for folder ordering** — must advance on *every* new thread, customer or staff.
2. **Wait-start anchor for the "Waiting Since" metric** — must *not* advance on consecutive customer messages once the customer is already waiting.

PR #5475 fixed (2) by breaking (1), because both read/write the same column.

## Fix

- `FetchEmails.php` / `ThreadObserver.php`: revert to unconditionally updating `last_reply_at`/`last_reply_from` on every thread (restores pre-#5475 behavior, fixes #5501).
- `Conversation.php`: add `getCustomerWaitingSince()`, which — only when the customer is the current last replier — walks the conversation's threads backwards from most recent and stops at the first non-customer reply, returning the timestamp of the oldest thread in that unbroken run. `getWaitingSince()` uses this instead of the raw `last_reply_at` column when applicable. `Folder::getWaitingSince()` (which *picks* the featured conversation) is untouched — only the *displayed* value for that conversation changes.

This keeps the two concerns fully decoupled: no schema change, no change to which conversation sorts where — only the "waiting since" text shown for a conversation where the customer is the last replier.

## Tested scenarios (same table as #5475)

| Scenario | Expected `Waiting Since` | Result |
|---|---|---|
| Two consecutive customer messages | Time of first message | ✅ |
| Three consecutive customer messages | Time of first message | ✅ |
| Customer → staff → customer | Time of last customer message | ✅ |
| First message (new conversation) | Message time | ✅ |
| Two consecutive staff messages | Time of last staff message | ✅ |
| Two consecutive customer messages, no staff reply | Conversation still sorts by latest activity in folder lists (#5501 no longer regresses) | ✅ |

Closes #5225

### DIFF
--- a/app/Console/Commands/FetchEmails.php
+++ b/app/Console/Commands/FetchEmails.php
@@ -1230,16 +1230,11 @@ class FetchEmails extends Command
         if ($conversation->status != Conversation::STATUS_ACTIVE && $conversation->status != Conversation::STATUS_SPAM) {
             $conversation->status = \Eventy::filter('conversation.status_changing', Conversation::STATUS_ACTIVE, $conversation);
         }
-        // This broke conversations order.
-        // https://github.com/freescout-help-desk/freescout/issues/5501#issuecomment-5084418616
-        //
-        // Only update last_reply_at if the customer was not already the last to reply.
-        // This preserves the original "waiting since" time when consecutive customer
-        // messages arrive without a staff reply in between.
-        // https://github.com/freescout-help-desk/freescout/issues/5225
-        /*if ($conversation->last_reply_from != Conversation::PERSON_CUSTOMER) {
-            $conversation->last_reply_at = $now;
-        }*/
+        // last_reply_at always reflects the true last activity time: it also drives
+        // folder list ordering (see Folder::getOrderByArray()), so it must advance on
+        // every incoming customer message, even consecutive ones without a staff reply
+        // in between (see #5501). The "waiting since" display is computed separately
+        // from thread history in Conversation::getCustomerWaitingSince() (see #5225).
         $conversation->last_reply_at = $now;
         $conversation->last_reply_from = Conversation::PERSON_CUSTOMER;
         // Reply from customer to deleted conversation should undelete it.
--- a/app/Conversation.php
+++ b/app/Conversation.php
@@ -1657,14 +1657,59 @@ class Conversation extends Model
         }
         $waiting_since_field = $folder->getWaitingSinceField();
         if ($waiting_since_field) {
-            // For phone conversations.
-            if (empty($this->$waiting_since_field)) {
-                $waiting_since_field = 'updated_at';
+            // last_reply_at reflects the time of the last activity of any kind (it also
+            // drives folder list ordering), not specifically how long the customer has
+            // been waiting for a reply. When the customer is the last person to have
+            // replied, compute the actual wait start from thread history instead so that
+            // consecutive customer messages don't keep resetting the displayed time
+            // (#5225), without touching last_reply_at itself (#5501).
+            if ($waiting_since_field == 'last_reply_at' && $this->last_reply_from == self::PERSON_CUSTOMER) {
+                $waiting_since_value = $this->getCustomerWaitingSince();
+            } else {
+                $waiting_since_value = $this->$waiting_since_field;
             }
-            return \App\User::dateDiffForHumans($this->$waiting_since_field);
+
+            // For phone conversations, and as a general fallback.
+            if (empty($waiting_since_value)) {
+                $waiting_since_value = $this->updated_at;
+            }
+            return \App\User::dateDiffForHumans($waiting_since_value);
         } else {
             return '';
         }
+    }
+
+    /**
+     * Get the timestamp from which the customer has been waiting for a reply,
+     * i.e. the creation time of the first thread in the current, unbroken run
+     * of customer messages. Unlike last_reply_at (which advances on every new
+     * thread so that folder lists keep sorting by true recency), this walks
+     * the conversation's threads backwards and stops as soon as it hits a
+     * reply that wasn't from the customer.
+     */
+    public function getCustomerWaitingSince()
+    {
+        $waiting_since = $this->last_reply_at;
+
+        $types = [Thread::TYPE_CUSTOMER, Thread::TYPE_MESSAGE];
+        if ($this->isPhone()) {
+            $types[] = Thread::TYPE_NOTE;
+        }
+
+        $threads = $this->threads()
+            ->whereIn('type', $types)
+            ->where('state', Thread::STATE_PUBLISHED)
+            ->orderBy('created_at', 'desc')
+            ->get(['source_via', 'created_at']);
+
+        foreach ($threads as $thread) {
+            if ($thread->source_via != self::PERSON_CUSTOMER) {
+                break;
+            }
+            $waiting_since = $thread->created_at;
+        }
+
+        return $waiting_since;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


freescout-help:
Thanks. But solution iterating over threads can not be accepted.

MarkoD:
@freescout-help The "waiting since" metric was mentioned as not being correct in multiple tickets (e.g., #5222, #5225 - both of them closed by the _now reverted_ commit), so I think this deserves a fix nonetheless.

How about introducing another variable (`last_customer_reply_at`)? That wouldn't break anything and could be updated properly alongside `last_reply_at` without iterating over threads.

freescout-help:
The thing is that there is no any good solution for this kind of issue:

1) If we'll skip updating `last_reply_at` field if previous reply belongs to the customer, we'll be getting this issue https://github.com/freescout-help-desk/freescout/issues/5501

2) If we'll use some extra `last_customer_reply_at` field to calculate "Waiting Since", we still will have to sort conversations in the list using `last_reply_at` because if using `last_customer_reply_at` we'll be getting https://github.com/freescout-help-desk/freescout/issues/5501. So when sorting using `last_reply_at` and using `last_customer_reply_at` to calculate "Waiting Since" in conversations list, "Waiting Since" value will be out of sorting order and we'll be receiving complaints that conversations are sorted in a wrong order.

ilyakonovalenko:
> 1. If we'll skip updating `last_reply_at` field if previous reply belongs to the customer, we'll be getting this issue [The conversation doesn't appear in the list when the customer replies several times #5501](https://github.com/freescout-help-desk/freescout/issues/5501)

It seems the sorting must be done by some last update time value, not last reply time value. So last update time changes on every conversation update (no matter who updates) so that sorting works well.

MarkoD:
@freescout-help I understand. We have 2 things that need to be addressed:

1. `last_reply_at` must be updated with each reply (the current state)
2. Because of the previous point, `last_reply_at` cannot be displayed in the UI as "Customer Waiting Since"
3. "Customer Waiting Since" is equal to the [time of the last customer's reply](https://github.com/freescout-help-desk/freescout/issues/5225#issuecomment-3924628122)

So the current "Customer Waiting Since" should be renamed to "Last Reply" (or something similar). A new variable should be introduced (e.g., `last_customer_reply_at`) and a new column (now correctly named "Customer Waiting Since") can be then added to the sorting table.

MarkoD:
@ilyakonovalenko You're right. Sorting is done via `last_reply_at`, which is updated with each reply (either customer's or agents'), and that's okay. The problem is that the value of `last_reply_at` is called "Customer Waiting Since" but should be called "Last Reply At" or something similar. And the actual "Customer Waiting Since" timestamp should be tracked separately and updated only when the customer replies.

MarkoD:
@freescout-help FYI, with 77 votes, this is currently the [4th most wanted feature request](https://feedback.userreport.com/25a3cb5f-e4bd-4470-b6f3-79fcfaa8e90f/#idea/432121).

freescout-help:
The thing is that people already got used that if a customer replies several times each reply pulls the conversation to the top in the list. So when we changed this behavior (to satisfy this feature request https://feedback.userreport.com/25a3cb5f-e4bd-4470-b6f3-79fcfaa8e90f/#idea/432121) immediately the following "bug report" appeared: https://github.com/freescout-help-desk/freescout/issues/5225

So some people prefer the way it is now and some want it to be different. So the solution can be a .env file parameter enabling the code of this commit https://github.com/freescout-help-desk/freescout/commit/caa5aac1c3ae4602faad7d9332f7e19664f23852

MarkoD:
@freescout-help Thanks for clarifying. The `.env` parameter is a good idea.

So the commit [@caa5ac](https://github.com/freescout-help-desk/freescout/commit/caa5aac1c3ae4602faad7d9332f7e19664f23852) does not break anything (in terms of bugs), just in the UI, conversations, in which the customer replies multiple times, won't be at the top of the list? (#5501)

Isn't anything else wired to `last_reply_at`? For example, the workflows module and it's conditions? If yes, then that would need to be checked whether it won't break when someone changes the new setting in their `.env` file.

losimo:
@MarkoD good questions, both worth nailing down properly rather than assuming. I audited every reference to `last_reply_at` directly against `master` to answer them with certainty.

**Reads — 2 consumers, both in core, nothing else:**
- Sort order: `Folder.php:149,154,166` (`order_by[] = ['last_reply_at' => 'desc']`), `Folder.php:385` (default sort column), `Conversation.php:2445` (default sort column), `Conversation.php:2532` / `ConversationsController.php:3196` (`orderBy('last_reply_at')`).
- Display: `Conversation.php:505-512`, a tooltip string. That's it.

**Writes — all unconditional, all on `master` today:** `ThreadObserver.php`, `FetchEmails.php`, `ConversationsController.php`, and `Conversation.php::forward()` each stamp `$conversation->last_reply_at = $now` on any new thread, agent or customer, no branching. The old #5475 logic that branched on reply author is still sitting there commented out in `FetchEmails.php`/`ThreadObserver.php`.

**Your question 1 — confirmed.** In this codebase `last_reply_at` drives exactly two things: folder sort order and one tooltip. Re-enabling `caa5aac1` behind a flag cannot touch anything else in core, because there is nothing else to touch.

**Your question 2 — @freescout-help this one's for you.** Workflows isn't in this repository: no `Modules/Workflows` anywhere in `freescout-help-desk/freescout`, and a full-repo search for `last_reply_at` returns only the files listed above. So nobody outside the FreeScout team can verify from source whether Workflows conditions read `last_reply_at`. That's not a "probably fine," it's a closed-source gap only you can close.

Net result: the `.env` flag is provably safe for everything in core. Whether it's safe overall comes down to that one confirmation.

freescout-help:
The only thing in Workflows that may not work as expected (when the flag will be enabled) - `Last Customer Reply` condition, it will miss customer replies sent one after another.

MarkoD:
@freescout-help Well that's a bummer, that condition is a commonly used one... Can you patch Workflows so that the `Last Customer Reply` will work as expected either way?

Any other official modules dependent on `last_reply_at`?

freescout-help:
https://github.com/freescout-help-desk/freescout/issues/5225#issuecomment-5150446812

PR #5603, opened 2026-08-27
Configurable persistent storage disk (S3 support) for attachments and avatars

mpdude:
Attachments and user/customer avatars are currently hardcoded to the local disk, with no way to point them at S3 (or an S3-compatible provider). This adds a `persistent_disk` config, defaulting
to today's local behavior, so both can be moved to durable/shared storage.

## Configuration

```
PERSISTENT_DISK=s3
AWS_ACCESS_KEY_ID=...
AWS_SECRET_ACCESS_KEY=...
AWS_DEFAULT_REGION=eu-central-1
AWS_BUCKET=freescout-persistent

# Only for S3-compatible non-AWS providers (MinIO, Hetzner Object Storage, ...)
AWS_ENDPOINT=
AWS_USE_PATH_STYLE_ENDPOINT=true
AWS_URL=
```

One disk for both attachments and avatars. `league/flysystem-aws-s3-v3` moved from `require-dev`
(where it already sat, unused) to `require`.

## Implementation notes

- No code assumes a local filesystem path anymore. Everywhere a real file or PHP resource is
  actually required by a third-party API (SwiftMailer attachments, GD image encoding,
  `UploadedFile`-based copy helpers), the bytes are streamed from Flysystem into a temp
  stream/file instead of relying on the disk having a native path.
- `Attachment::getLocalFilePath()` is kept for the local-disk case but no longer used at the three
  spots that broke under S3 (reply attachments, conversation-forward attachment copy, avatar
  encoding).

## Known limitation

No migration command for moving *existing* local attachments/avatars into a newly-configured
bucket - only newly-written files go through the configured disk. Existing files need a manual
copy (e.g. `aws s3 sync`) if switching an already-running instance.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,3 +51,21 @@ APP_KEY=
 
 # Uncomment to see errors in your browser, don't forget to comment it back when debugging finished
 #APP_DEBUG=true
+
+# Disk used to store conversation attachments and user/customer avatars. Defaults to
+# "private" (local disk, storage/app). Set to "s3" to store this data on S3 (or any
+# S3-compatible provider) instead, e.g. for horizontally-scaled deployments.
+#PERSISTENT_DISK=s3
+
+# S3 credentials, only needed when PERSISTENT_DISK=s3.
+#AWS_ACCESS_KEY_ID=
+#AWS_SECRET_ACCESS_KEY=
+#AWS_DEFAULT_REGION=
+#AWS_BUCKET=
+
+# Only needed for S3-compatible providers other than AWS itself (e.g. MinIO, Hetzner
+# Object Storage, Wasabi, DigitalOcean Spaces).
+#AWS_ENDPOINT=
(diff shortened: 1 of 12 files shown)


freescout-help:
FYI. The first issue we've faced is the size of `league/flysystem-aws-s3-v3` package - it brings >70Mb of code while current FreeScout's unpacked dist size is 108Mb.

So we'll try to implement S3 storage support via the module - we'll send the link to the GitHub repo.

mpdude:
@freescout-help You mean you're going to invite me to a dedicated repo where you develop S3 support as a module?

freescout-help:
https://github.com/freescout-help-desk/freescout/commit/12adf2470cd76b0f4e4b27f0b26a80cf9bd267fb

https://github.com/freescout-help-desk/S3Storage

Issue #5591, opened 2026-08-24
FreeScout silently skips legitimate emails when an external sender reuses the same Message-ID


Kerovsky:
<h2>FreeScout silently skips legitimate emails when an external sender reuses the same Message-ID</h2>

<p>We encountered an issue where FreeScout can silently skip legitimate incoming emails if an external mail system sends multiple different messages using exactly the same <code>Message-ID</code>.</p>

<p>In our case, two completely different emails were received on the same IMAP mailbox:</p>

<pre><code>First email:
Date: 2026-06-25
Message-ID: 3klmoZ1mRr8MHpDmsEMpr71782370995962@mct.prg-dc.dhl.com

Second email:
Date: 2026-07-03
Message-ID: 3klmoZ1mRr8MHpDmsEMpr71782370995962@mct.prg-dc.dhl.com
</code></pre>

<p>The emails had different dates and different content, but the external sender reused the same <code>Message-ID</code>.</p>

<p>FreeScout fetched the second message from IMAP, but then rejected it with:</p>

<pre><code>Message with such Message-ID has been fetched before:
3klmoZ1mRr8MHpDmsEMpr71782370995962@mct.prg-dc.dhl.com
</code></pre>

<p>The message was not added to the conversation and was additionally marked as <code>Seen</code> on the IMAP server.</p>

<p>The problem appears to come from the duplicate check, which treats an existing <code>Message-ID</code> as sufficient proof that the incoming message has already been fetched.</p>

<p>In practice, not all third-party systems generate unique Message-IDs correctly. Ticketing systems, courier systems, ERP systems or automated mail gateways may sometimes reuse the same <code>Message-ID</code> for different physical emails.</p>

<p>This creates a serious problem because:</p>

<ul>
    <li>the email physically exists on the IMAP server,</li>
    <li>FreeScout detects it,</li>
    <li>FreeScout rejects it as a duplicate,</li>
    <li>the message is marked as read,</li>
    <li>the email never appears in FreeScout,</li>
    <li>the user may not notice that an important message was lost.</li>
</ul>

<p>I think FreeScout should distinguish between:</p>

<ol>
    <li>the same email being fetched again,</li>
    <li>a different email where the external sender reused the same <code>Message-ID</code>.</li>
</ol>

<p>When a duplicate <code>Message-ID</code> is found, FreeScout could additionally compare values such as:</p>

<pre><code>Date
From
To
Subject
IMAP UID
message size
body/content hash
</code></pre>

[...]

Issue #5434, opened 2026-05-28
Integrity constraint violation on multiple receivers


sofl88:
Hello!

When I try to send an email from one freescout mailbox to two others (both in the "to" field), I get this error when fetching

```
Error: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'fs-a8c2bf3f706cf31af5f33e3e5c6443bc@xxxx.de' for key 'threads_message_id_index' (SQL: insert into `threads` (`conversation_id`, `user_id`, `type`, `status`, `state`, `message_id`, `headers`, `body`, `from`, `to`, `cc`, `bcc`, `source_via`, `source_type`, `customer_id`, `created_by_customer_id`, `created_at`, `updated_at`) values (
```

Thanks in advance

PHP version: 8.3.29
FreeScout version: 1.8.218
Database: 10.11.14-MariaDB
Are you using CloudFlare: No
Are you using non-official modules: Yes

freescout-help:
We can't reproduce the issue. Can you send a screenshot of how your are sending the email.

sofl88:
Ok, I have these 3 mailboxes.

<img width="865" height="306" alt="Image" src="https://github.com/user-attachments/assets/8296c032-c22b-42e8-9d76-7c599f35276f" />

Now I want to send an email from "Catch all" to "Test" and "Info"

<img width="1241" height="333" alt="Image" src="https://github.com/user-attachments/assets/d6aad8d3-ccea-4497-b6a9-1ae2c5d9e6ca" />

Now if everything is sent or after I forced it with `php artisan queue:work --queue=emails,default`, I manually run the fetch process

<img width="1205" height="394" alt="Image" src="https://github.com/user-attachments/assets/edf9bd23-ac9f-4e8d-8c38-3b750bff8246" />

Thanks in advance

freescout-help:
We can't reproduce it. The `$mailbox->id` is different on this line, so generated Message ID can't be the same:

https://github.com/freescout-help-desk/freescout/blob/dist/app/Console/Commands/FetchEmails.php#L440

sofl88:
Okay thanks for your effort maybe I've missed something here.

sofl88:
[...]

PR #5636, opened 2026-09-15
Prevent duplicate multi-mailbox imports

### LINKED ISSUE #5485: Duplicate imports into target mailbox when sending from one FreeScout mailbox to another FreeScout mailbox with a distribution list in CC
PHP version: 8.4.21
FreeScout version:  1.8.223
Database: MySQL
Mail server: Zimbra
Are you using CloudFlare: No
Are you using non-official modules: No

**Summary**

We are seeing duplicate conversations/messages created in one FreeScout mailbox when an email is sent from one FreeScout mailbox to another FreeScout mailbox, with a Zimbra distribution list in CC.

This does not appear to be an SMTP resend issue. FreeScout sends the email once, but during fetching/importing it creates many artificial fs-* Message-IDs and repeatedly imports the same logical email into the target mailbox.

This looks related to the multi-mailbox import logic in app/Console/Commands/FetchEmails.php.

Possibly related to issues such as  #5368 and #5434, but the trigger here is a distribution list / alias in CC that expands to many addresses, some of which are also FreeScout-monitored mailboxes.

**Setup**

We have several FreeScout mailboxes on the same instance, for example:

Mailbox A: communication@example.org
Mailbox B: cospc@example.org
Other FreeScout mailboxes:
- animation@example.org
- police@example.org
- accueil@example.org
- scolaire@example.org
- comptabilite@example.org
- etc.

The mail server is Zimbra.

There is a Zimbra distribution list / alias:

vdb-services@example.org

This distribution list contains many recipients, including several addresses that are also configured as FreeScout mailboxes.

**Scenario**

From FreeScout mailbox A:

From: communication@example.org
To: cospc@example.org
Cc: vdb-services@example.org

The distribution list expands on Zimbra to many recipients. Several of those recipients are mailboxes monitored by FreeScout.

**Expected behavior**

FreeScout should create/import the email once in the target mailbox:

cospc@example.org

Other FreeScout mailboxes that physically received the message via the distribution list may also import their own copy if appropriate, but they should not cause the target mailbox to receive a new duplicate import each time.

**Actual behavior**

FreeScout sent the email only once, but the target mailbox received/imported the same logical email 21 times in about one minute.

The outgoing send log shows only one send attempt:

```
send_logs:
thread_id: 138459
message_id: FS_reply-138459-d09b06a1eed657dd@example.org
recipient: cospc@example.org
status: accepted

thread_id: 138459
message_id: FS_reply-138459-d09b06a1eed657dd@example.org
recipient: vdb-services@example.org
status: accepted
```

No repeated SMTP sends were logged.

However, the threads table shows the following import pattern:

```
14:11:06  Service Animation   FS_reply-138459-d09b06a1eed657dd@example.org
14:11:06  COSPC               fs-5774171721687dd30b3bb378f2093f41@example.org

14:11:08  Service Communication fs-6ae4c19c6f94652607b1094e75f63ed0@example.org
14:11:08  COSPC                 fs-783d8f4458d603c5442d353185069feb@example.org

14:11:10  Service de Police   fs-88807acab8e96fa294a6e194d9c9931c@example.org
14:11:10  COSPC               fs-95af28c07c90b327b0b98a714101db68@example.org

14:11:15  Accueil             fs-f079bab5b15b77e08bbf22f19271aef9@example.org
14:11:16  COSPC               fs-60d075c7b29c6b9440c65c3d4629c32d@example.org

...
```

Grouped by mailbox:

```
COSPC mailbox:
  thread_count: 21
  original FS_reply count: 0
  artificial fs-* Message-ID count: 21
  first seen: 14:11:06
  last seen: 14:12:11

Several other FreeScout mailboxes:
  thread_count: 1 each
  artificial fs-* Message-ID count: 1 each
```

All these imported messages had the visible headers stored in FreeScout as:

```
From: communication@example.org
To: ["cospc@example.org"]
Cc: ["vdb-services@example.org"]
```

So every physical copy delivered through the distribution list still visibly targets cospc@example.org.

Suspected root cause

In app/Console/Commands/FetchEmails.php, FreeScout has special logic to import an email into other FreeScout mailboxes when several FreeScout mailbox addresses are present in the visible recipients.

The relevant part builds $recipient_emails from visible To, Cc, and Bcc headers:

```
$recipient_emails = array_unique($this->formatEmailList(array_merge(
    $this->attrToArray($message->getTo()), 
    $this->attrToArray($message->getCc()), 
    $this->attrToArray($message->getBcc())
)));
```

Then it loops over all FreeScout mailboxes and adds an extra_import if the mailbox email appears in those visible recipients:

```
foreach ($mailboxes as $check_mailbox) {
    ...
    foreach ($recipient_emails as $recipient_email) {
        // No need to check mailbox aliases.
        if (\App\Email::sanitizeEmail($check_mailbox->email) == $recipient_email) {
            $this->extra_import[] = [
                'mailbox'    => $check_mailbox,
                'message'    => $message,
                'message_id' => $message_id,
            ];
            break;
        }
    }
}
```

When processing an extra import, FreeScout generates an artificial Message-ID:

```
$message_id = \MailHelper::generateMessageId(
    strstr($message_id, '@') ? $message_id : $from,
    $mailbox->id.$message_id
);
```

In our case, the distribution list causes many FreeScout mailboxes to physically receive a copy of the same logical email. Each copy still has:

```
To: cospc@example.org
Cc: vdb-services@example.org
```

So for each mailbox fetched through the distribution list, FreeScout appears to decide:

This email also targets cospc@example.org
→ create an extra import into the COSPC mailbox
→ generate a new artificial fs-* Message-ID

Because a new artificial fs-* Message-ID is generated, normal message_id deduplication does not stop the repeated imports into the same target mailbox.

AdamCoffeeOverflow:
## What changed

Keep the pre-mailbox-specific Message-ID available when an email has already been imported into another mailbox and is being processed as an extra mailbox copy.

When that copy causes another `extra_import`, pass the preserved Message-ID instead of the already generated `fs-*` ID. Also skip scheduling an extra import back into the mailbox that already owns the original Message-ID.

## Why

In #5485, a distribution list delivers the same email to several FreeScout-monitored mailboxes while the visible headers still contain the target FreeScout mailbox.

Each physical copy can be treated as an extra import and receive a mailbox-specific `fs-*` Message-ID. That modified ID was then passed into the target mailbox's `extra_import`, so each source mailbox produced a different artificial Message-ID in the target mailbox.

Because those target IDs were different, the existing #5434 duplicate check could not catch them.

With this change, physical copies of the same message use the same pre-mailbox-specific Message-ID when creating the downstream extra import. The target mailbox therefore generates the same artificial Message-ID, allowing the existing duplicate check to stop subsequent imports.

The existing behavior for genuinely different emails reusing the same Message-ID (#5591) is left unchanged.

## Checks

* `php -l app/Console/Commands/FetchEmails.php`
* `git diff --check`
* Branch based directly on current `master`
* Final diff changes only `app/Console/Commands/FetchEmails.php`

Fixes #5485


freescout-help:
Thanks. `message_id` field is supposed to be unique in `threads` table.

Also https://github.com/freescout-help-desk/freescout/issues/5485#issuecomment-5692994044

PR #5641, opened 2026-09-17
fix: only serve an attachment from disk when the disk is local

jhgeluk:
Viewable attachments with BinaryFileResponse
so audio and video get HTTP Range requests. That reads the local filesystem, which has no meaning on a remote disk. So Storage::path() hands back a storage-relative path and Symfony's File throws FileNotFoundException on every image and PDF.

Fall back to Storage::download(), which streams the blob. That costs Range support, so audio and video can not be seeked while attachments live on S3 or Azure.

This allows a custom fileadapter for S3 or Azure to be used, seems to be broken since `1.8.237`


freescout-help:
Maybe there is some more elegant way to determine that non-local filesystem is used.
What values did you set in .env file to enable S3 storage?

jhgeluk:
> Maybe there is some more elegant way to determine that non-local filesystem is used. What values did you set in .env file to enable S3 storage?

Hi, What would you propose? We're currently not using S3 but Azure blob storage, so made our own file adapter for this. So we have our own custom environment variables. The cleanest way in my opinion would be just checking if it's using the local file adapter

freescout-help:
Can you use something like `config("filesysmtes.default") == "local"` ?

freescout-help:
Can you share the content of your .env file related to AWS S3?

PR #5651, opened 2026-09-19
Trim trailing whitespace in User::getFullName()

il2020:
### Problem

Emails sent to customers by the Workflows module render a stray space before the comma in the signature:

```
--
Служба поддержки ,
```

Root cause chain:

1. The Workflows module creates its robot user by putting the whole `WORKFLOWS_USER_FULL_NAME` value into `first_name` and leaving `last_name` empty (`Modules/Workflows/Entities/Workflow.php:1654-1657`).
2. The `{%user.fullName%}` merge code in signatures resolves via `$data['user']->getFullName()` (`app/Misc/Mail.php:320`).
3. `User::getFullName()` concatenated unconditionally — `first_name . ' ' . last_name` — so an empty last name yielded `"Служба поддержки "`, and a signature template of `{%user.fullName%},` produced `Служба поддержки ,`.

This isn't specific to Workflows: any user without a last name gets a trailing space wherever their full name is displayed.

### Fix

Wrap the concatenation in `trim()` in `app/User.php:235`. This matches the behavior already implemented in `Customer::getFullName()` (`app/Customer.php:526-534`), which handles the empty-last-name case explicitly. The value is passed to the `user.full_name` Eventy filter already trimmed, so filter subscribers receive a clean string too.

```diff
-return \Eventy::filter('user.full_name', $this->first_name.' '.$this->last_name, $this);
+return \Eventy::filter('user.full_name', trim($this->first_name.' '.$this->last_name), $this);
```

### Scope / side effects

The change is global by design: anywhere a user with an empty last name is rendered (user lists, assignee names, sender names, mail merge codes) loses the trailing space. No call site relies on the trailing space.


freescout-help:
Thanks. It looks like some extra commits were added to this PR.
Added as https://github.com/freescout-help-desk/freescout/pull/5651/changes/11566aaf7bc96fddbd33f1a2f639f3b978dbe620

